PR #106, opened 2020-04-16
Make alpha optional when parsing a Color from an input stream

azeey:
The stream insertion operator is used in libsdformat to parse Color values, and as far as I know, alpha is already optional, but currently, if the exception mask for [`failbit`](https://en.cppreference.com/w/cpp/io/basic_ios/exceptions) is set ([we are doing that in libsdformat to catch more parsing errors](https://github.com/osrf/sdformat/pull/244)), the `operator>>` function will throw an exception. This PR keeps the current behavior by keeping alpha optional, but prevents an exception from being thrown.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,9 @@
 
 ## Ignition Math 6.x.x
 
+1. Make alpha optional when parsing a Color from an input stream.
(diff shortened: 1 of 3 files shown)


j-rivero:
@osrf-jenkins run tests

iche033:
update changelog?

changes look good to me

Issue #35, opened 2015-12-05
Add setter/getter for Pose's each element


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-math/issues/35)) by Louise Poubel (Bitbucket: [chapulina](https://bitbucket.org/%7B5cfa2075-477b-4ded-bdb9-8d2479544ec4%7D/), GitHub: [chapulina](https://github.com/chapulina)).**

----------------------------------------
It would be helpful to be able to do `pose.X()` instead of `pose.Pos.X()`.

How about this?

* `pose.Pos.X()` -> `pose.X()`
* `pose.Pos.Y()` -> `pose.Y()`
* `pose.Pos.Z()` -> `pose.Z()`
* `pose.Rot.Roll()` -> `pose.Roll()`
* `pose.Rot.Pitch()` -> `pose.Pitch()`
* `pose.Rot.Yaw()` -> `pose.Yaw()`

pxalcantara:
Currently I'm working on this issue!

Issue #101, opened 2019-08-07
Documentation, Style and scripting interfaces update


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-math/issues/101)) by Nate Koenig (Bitbucket: [Nathan Koenig](https://bitbucket.org/%7Bc862cdd9-fcc8-4419-9fc7-e20db14b8fcb%7D/)).**

----------------------------------------
# Summary

The documentation,  code style, and scripting interfaces in this library needs to be updated. This issue will track progress.

# Work to be done

For each class:

1. Create an example in the `examples` folder. This example should then be used as a doxygen `\snippet` in the class's doxygen description.

2. Check spelling.

3. Check for code style. Set functions should start with `Set`. Don't use `const` with POD types. Etc.

4. Create a scripting interface. See Angle.i as an example.

5. Add more tests so that  coverage is as close to 100% as possible.

6. Target changes to ign-math6. Also, target `main` with any breaking changes.

# Status

1. [x] Angle. See [pull request #325](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-math/pull-requests/325) and [pull request #326](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-math/pull-requests/326). https://github.com/ignitionrobotics/ign-math/pull/216
2. [x] AxisAlignedBox https://github.com/ignitionrobotics/ign-math/pull/262 https://github.com/ignitionrobotics/ign-math/pull/261
3. [ ] Box https://github.com/ignitionrobotics/ign-math/pull/273
4. [x] Color https://github.com/ignitionrobotics/ign-math/pull/218
5. [x] Cylinder https://github.com/ignitionrobotics/ign-math/pull/274
6. [ ] DiffDriveOdometry https://github.com/ignitionrobotics/ign-math/pull/265
7. [x] Filter https://github.com/ignitionrobotics/ign-math/pull/230
8. [ ] Frustum https://github.com/ignitionrobotics/ign-math/pull/278
9. [x] GaussMarkovProcess https://github.com/ignitionrobotics/ign-math/pull/216
10. [ ] graph/Edge
11. [ ] graph/GraphAlgorithms
12. [ ] graph/Graph
13. [ ] graph/Vertex
14. [x] Helpers https://github.com/ignitionrobotics/ign-math/pull/266
[...]

PR #125, opened 2020-06-06
Add setter/getter for Pose's each element - Issue 35

pxalcantara:
Solved #35 

Add methods to Pose3.hh to get X,Y,Z,Roll,Pitch and Yaw directly, as suggested in the issue description : 

- pose.Pos.X() -> pose.X()
- pose.Pos.Y() -> pose.Y()
- pose.Pos.Z() -> pose.Z()
- pose.Rot.Roll() -> pose.Roll()
- pose.Rot.Pitch() -> pose.Pitch()
- pose.Rot.Yaw() -> pose.Yaw()

### DIFF
--- a/src/Pose_TEST.cc
+++ b/src/Pose_TEST.cc
@@ -141,7 +141,7 @@ TEST(PoseTest, Pose)
 TEST(PoseTest, ConstPose)
 {
   const math::Pose3d pose(0, 1, 2, 1, 0, 0);
-  
+
   EXPECT_TRUE(pose.Pos() == math::Vector3d(0, 1, 2));
   EXPECT_TRUE(pose.Rot() == math::Quaterniond(1, 0, 0));
 }
@@ -168,13 +168,12 @@ TEST(PoseTest, MutablePose)
 
   EXPECT_TRUE(pose.Pos() == math::Vector3d(10, 20, 30));
   EXPECT_TRUE(pose.Rot() == math::Quaterniond(1, 2, 1));
- }
+}
 
 /////////////////////////////////////////////////
 TEST(PoseTest, ConstPoseElements)
 {
   const math::Pose3d pose(0, 1, 2, 1, 1, 2);
- 
   EXPECT_DOUBLE_EQ(pose.X(), 0);
   EXPECT_DOUBLE_EQ(pose.Y(), 1);
   EXPECT_DOUBLE_EQ(pose.Z(), 2);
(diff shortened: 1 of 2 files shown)


pxalcantara:
@chapulina Here we go! first PR!! \o/

src/Pose_TEST.cc
```diff
@@ -155,3 +155,22 @@ TEST(PoseTest, OperatorStreamOut)
...
+TEST(PoseTest, PoseElements)
+{
+  const math::Pose3i pose(0, 1, 2, 0, 0, 0);
+
+  int x = pose.Pos().X();
+  int y = pose.Pos().Y();
+  int z = pose.Pos().Z();
+  int Roll = pose.Rot().Roll();
+  int Pitch = pose.Rot().Pitch();
+  int Yaw = pose.Rot().Yaw();
```
chapulina, line 168: Small style detail: could you use lowercase for all variables? Such as:

```suggestion
  int yaw = pose.Rot().Yaw();
```

pxalcantara: ohh yeah, sure, I can do this. I've forget about the variable style code, sorry!

scpeters:
are you able to merge this branch with `ign-math6`? it is currently out of date with the base branch

pxalcantara:
[DO NOT MERGE] @chapulina @scpeters 
This code is not finished. The approach used to implement the mutable X(), Y() and Z() can not be used in Roll(), Pitch() and Yaw(). When this approach is used I got these errors:

RPY_error

That is expected because the Quaternion.Roll() , Pitch() and Yaw() are `const` : https://github.com/ignitionrobotics/ign-math/blob/ca49e983bc3939b6626e78f036ff5ea9bf10d270/include/ignition/math/Quaternion.hh#L423

How should I solve this point? I thought in some ways but, all of them I need to edit the library Quaternion.hh and, also, there is pattern and architecture questions, so I decided to ask

chapulina:
humm nice catch

In fact, this reminds me that we've been deprecating the mutable accessors in favor of `Set*` functions, see the [Migration guide](https://github.com/ignitionrobotics/ign-math/blob/master/Migration.md) and issue #101 .

So I think the solution is to substitute the mutable functions with `Set` versions.

pxalcantara:
Humm, nice, so, maybe, stop working with this issue and help with this functionality of the Quaternio.hh ? After that continue to working here?

chapulina:
Oh I don't think you need to address the entire issue. I'd just change the new functions you're adding. Remove all the mutable functions and add:

```
Pose3d::SetX
Pose3d::SetY
Pose3d::SetZ
Pose3d::SetRoll
Pose3d::SetPitch
Pose3d::SetYaw
```

pxalcantara:
Ok, good, should I remove Pos() and Rot() mutable, as well, and change for SetPos() and SetRot(), or is it better to do this in another PR?

chapulina:
> should I remove Pos() and Rot() mutable, as well, and change for SetPos() and SetRot()

Good call, yes, please!

scpeters:
> Ok, good, should I remove Pos() and Rot() mutable, as well, and change for SetPos() and SetRot(), or is it better to do this in another PR?

those API's are already released in ign-math6, so I wouldn't remove them in this PR

chapulina:
> those API's are already released in ign-math6, so I wouldn't remove them in this PR

Ah sorry, thank you Steve. Let's only remove the ones being added by this PR.

scpeters:
> This code is not finished. The approach used to implement the mutable X(), Y() and Z() can not be used in Roll(), Pitch() and Yaw().

Quaternion doesn't allow modification of Euler angles independently; it only allows them to all be set at once. I would suggest simply adding `Get` functions for roll, pitch, yaw in this PR and address setting them in a separate pull request

pxalcantara:
> > This code is not finished. The approach used to implement the mutable X(), Y() and Z() can not be used in Roll(), Pitch() and Yaw().
> 
> Quaternion doesn't allow modification of Euler angles independently; it only allows them to all be set at once. I would suggest simply adding `Get` functions for roll, pitch, yaw in this PR and address setting them in a separate pull request

Ok, It was I thought, and also using the Euler method would not be a good idea now since it is deprecated, so I'll do your suggestion!
Thanks @scpeters and @chapulina

pxalcantara:
@scpeters @chapulina sorry for the delay. Is this the way that you suggested? About the failure of the CI test, I couldn't understand the error, what should I do to fix it?

scpeters:
it looks like the build machines had an error; I'll start the jobs again

scpeters:
@osrf-jenkins run tests please

pxalcantara:
Thank you @scpeters!

PR #132, opened 2020-07-13
Vector3 to vector4 functions

### LINKED ISSUE #71: Add functions from Vector3 to Vector2 and Vector4
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-math/issues/71)) by Louise Poubel (Bitbucket: [chapulina](https://bitbucket.org/%7B5cfa2075-477b-4ded-bdb9-8d2479544ec4%7D/), GitHub: [chapulina](https://github.com/chapulina)).**

For example, there is a [Vector3::Normalized](https://github.com/ignitionrobotics/ign-math/blob/05458c89f0434e2774f19995e713ed7dd94f1018/include/ignition/math/Vector3.hh#L143), but no such function for [Vector2](https://github.com/ignitionrobotics/ign-math/blob/05458c89f0434e2774f19995e713ed7dd94f1018/include/ignition/math/Vector2.hh) or [Vector4](https://github.com/ignitionrobotics/ign-math/blob/05458c89f0434e2774f19995e713ed7dd94f1018/include/ignition/math/Vector4.hh). The same goes for `Round`, `Rounded` and other functions.

IssueForNewDevelopers

luccosta:
Resolves #71 .

This implements the functions Min(), Min(Vector4), Max() and Max(Vector4) in Vector4 API, similar to the existing same functions in Vector3.

It enables Vector4 objects to perform:

`
vec4_a.max(vec4_b);
vec4_a.min(vec4_b);
vec4.max();
vec4.min();
`

All the tests passed, and I dont know how to evaluate the code coverage.

To build the changes correctly i had to adjust some files. 

One was src/CMakeLists.txt to add Vector4 to the swig files, because the include of Vector4 library at Vector4_TEST.cc was using build/fake/install/include/ignition/math7/ignition/math/Vector4.hh. Also had to change RollingMean.hh because the lack of the limits library was causing build errors.

This is my first pull request, I was in doubt if was better make the PR with this adjusts asking if they were correct or ask about them before make the PR and if the correct thing to do is to make this adjusts in another PR. Also, I would like to receive advices to improvement.

### DIFF
--- a/include/ignition/math/RollingMean.hh
+++ b/include/ignition/math/RollingMean.hh
@@ -18,7 +18,6 @@
 #define IGNITION_MATH_ROLLINGMEAN_HH_
 
 #include <memory>
-#include <limits>
 #include <ignition/math/Export.hh>
 #include <ignition/math/config.hh>
 
--- a/include/ignition/math/RollingMean.hh
+++ b/include/ignition/math/RollingMean.hh
@@ -18,6 +18,7 @@
 #define IGNITION_MATH_ROLLINGMEAN_HH_
 
 #include <memory>
+#include <limits>
 #include <ignition/math/Export.hh>
 #include <ignition/math/config.hh>
 
--- a/include/ignition/math/Vector4.hh
+++ b/include/ignition/math/Vector4.hh
@@ -133,14 +133,10 @@ namespace ignition
       /// \param[in] _v the maximum clamping vector
       public: void Max(const Vector4<T> &_v)
       {
-        if (_v[0] > this->data[0])
-          this->data[0] = _v[0];
-        if (_v[1] > this->data[1])
-          this->data[1] = _v[1];
-        if (_v[2] > this->data[2])
-          this->data[2] = _v[2];
-        if (_v[3] > this->data[3])
-          this->data[3] = _v[3];
+        this->data[0] = std::max(_v[0], this->data[0]);
+        this->data[1] = std::max(_v[1], this->data[1]);
+        this->data[2] = std::max(_v[2], this->data[2]);
+        this->data[3] = std::max(_v[3], this->data[3]);
       }
 
       /// \brief Set this vector's components to the minimum of itself and the
@@ -148,34 +144,24 @@ namespace ignition
       /// \param[in] _v the minimum clamping vector
       public: void Min(const Vector4<T> &_v)
       {
-        if (_v[0] < this->data[0])
-          this->data[0] = _v[0];
-        if (_v[1] < this->data[1])
-          this->data[1] = _v[1];
-        if (_v[2] < this->data[2])
-          this->data[2] = _v[2];
-        if (_v[3] < this->data[3])
-          this->data[3] = _v[3];
+        this->data[0] = std::min(_v[0], this->data[0]);
+        this->data[1] = std::min(_v[1], this->data[1]);
+        this->data[2] = std::min(_v[2], this->data[2]);
+        this->data[3] = std::min(_v[3], this->data[3]);
       }
 
       /// \brief Get the maximum value in the vector
       /// \return the maximum element
       public: T Max() const
       {
-        return std::max(std::max(std::max(this->data[0],
-                                          this->data[1]),
-                                          this->data[2]),
-                                          this->data[3]);
+        return *std::max_element(this->data, this->data+4);
       }
 
       /// \brief Get the minimum value in the vector
       /// \return the minimum element
       public: T Min() const
       {
-        return std::min(std::min(std::min(this->data[0],
-                                          this->data[1]),
-                                          this->data[2]),
-                                          this->data[3]);
+        return *std::min_element(this->data, this->data+4);
       }
 
       /// \brief Assignment operator
--- a/include/ignition/math/Vector4.hh
+++ b/include/ignition/math/Vector4.hh
@@ -17,6 +17,8 @@
 #ifndef IGNITION_MATH_VECTOR4_HH_
 #define IGNITION_MATH_VECTOR4_HH_
 
+#include <algorithm>
+
 #include <ignition/math/Matrix4.hh>
 #include <ignition/math/Helpers.hh>
 #include <ignition/math/config.hh>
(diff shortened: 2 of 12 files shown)


luccosta:
_ignition_math-abichecker-any_to_any-ubuntu_auto-amd64_ and _ignition_math-ci-pr_any-ubuntu_auto-amd64_ do not pass because they depends on the "Vector4.i" file, that I suppose to be an interface to ruby test code, and both does not exist on branch ign-math6 for Vector4.

I have started to write both "Vector4.i" and "Vector4_TEST.rb", the ideal is report this in an new Issue and make a new PR?

mjcarroll:
> ignition_math-abichecker-any_to_any-ubuntu_auto-amd64 and ignition_math-ci-pr_any-ubuntu_auto-amd64 do not pass because they depends on the "Vector4.i" file, that I suppose to be an interface to ruby test code, and both does not exist on branch ign-math6 for Vector4.

I believe if you want you can leave `Vector4` out of the list of `swig_files` here: https://github.com/ignitionrobotics/ign-math/pull/132/files#diff-95e351a3805a1dafa85bf20b81d086e6R30, which should cause that generation to be skipped.  We can then ticket that separately if you don't have it implemented right this second.

include/ignition/math/Vector4.hh
```diff
@@ -126,6 +128,56 @@ namespace ignition
...
+          this->data[2] = _v[2];
+        if (_v[3] < this->data[3])
+          this->data[3] = _v[3];
+      }
+
+      /// \brief Get the maximum value in the vector
+      /// \return the maximum element
+      public: T Max() const
+      {
+        return std::max(std::max(std::max(this->data[0],
```
mjcarroll, line 165: I believe that `std::max_element` would be a bit more concise: https://en.cppreference.com/w/cpp/algorithm/max_element

include/ignition/math/Vector4.hh
```diff
@@ -126,6 +128,56 @@ namespace ignition
...
+                                          this->data[1]),
+                                          this->data[2]),
+                                          this->data[3]);
+      }
+
+      /// \brief Get the minimum value in the vector
+      /// \return the minimum element
+      public: T Min() const
+      {
+        return std::min(std::min(std::min(this->data[0],
```
mjcarroll, line 175: Same as above but with `std::min_element`.

include/ignition/math/Vector4.hh
```diff
@@ -126,6 +128,56 @@ namespace ignition
         this->data[3] = _w;
       }
 
+      /// \brief Set this vector's components to the maximum of itself and the
+      ///        passed in vector
+      /// \param[in] _v the maximum clamping vector
+      public: void Max(const Vector4<T> &_v)
+      {
+        if (_v[0] > this->data[0])
+          this->data[0] = _v[0];
```
mjcarroll, line 137: I believe this would be more concise

```suggestion
        this->data[0] = std::max(_v[0], this->data[0]);
```

include/ignition/math/Vector4.hh
```diff
@@ -126,6 +128,56 @@ namespace ignition
...
+        if (_v[2] > this->data[2])
+          this->data[2] = _v[2];
+        if (_v[3] > this->data[3])
+          this->data[3] = _v[3];
+      }
+
+      /// \brief Set this vector's components to the minimum of itself and the
+      ///        passed in vector
+      /// \param[in] _v the minimum clamping vector
+      public: void Min(const Vector4<T> &_v)
```
mjcarroll, line 149: Same as above, but with `std::min`

src/Vector4_TEST.cc
```diff
@@ -20,6 +20,7 @@
 #include "ignition/math/Helpers.hh"
 #include "ignition/math/Matrix4.hh"
 #include "ignition/math/Vector4.hh"
+#include "../include/ignition/math/Vector4.hh"
```
mjcarroll, line 23: This doesn't seem necessary, the same header should be included above.

include/ignition/math/RollingMean.hh
```diff
@@ -18,6 +18,7 @@
 #define IGNITION_MATH_ROLLINGMEAN_HH_
 
 #include <memory>
+#include <limits>
```
mjcarroll, line 21: Out of scope for this PR?

luccosta: I agree and will remove from the PR, only put this library because before it the build is giving the following error:

`
ign-math/src/RollingMean.cc:57:10: error: 'numeric_limits' is not a member of 'std'
` 

Must I create an Issue for this?

src/Vector4_TEST.cc
luccosta: Ops, I forgot that rs, will remove

include/ignition/math/Vector4.hh
luccosta: I thing the same, and prefer `std::min`, only did with the ifs because is implemented this way in Vector3. Despite that I can use `std::min`?

include/ignition/math/Vector4.hh
luccosta: Much better.

luccosta:
> I believe if you want you can leave `Vector4` out of the list of `swig_files` here: https://github.com/ignitionrobotics/ign-math/pull/132/files#diff-95e351a3805a1dafa85bf20b81d086e6R30, which should cause that generation to be skipped. We can then ticket that separately if you don't have it implemented right this second.

Remove `Vector4` from `swig_files` and made the suggested changes. 

Also had finished `Vector4.i` and am close to finish `Vector4_TEST.rb`, that will be a relative big pull request.

include/ignition/math/Vector4.hh
mjcarroll: I prefer this, it would actually probably make sense to go back and fix vector3 and vector2 to be the same, but it's out of the scope of this pr.

include/ignition/math/RollingMean.hh
mjcarroll: It's fine to put in here, but since it's used in the cc file, rather than the header, it should be added there, I believe.

PR #139, opened 2020-07-20
Update actions CI

chapulina:
Update actions CI to work with https://github.com/ignition-tooling/ubuntu-bionic-ci-action/pull/13.

Also, show the collapsibles from https://github.com/ignition-tooling/ubuntu-bionic-ci-action/pull/14 in action.


.github/workflows/ci-bionic.yml
```diff
@@ -13,6 +13,9 @@ jobs:
         id: ci
         uses: ignition-tooling/ubuntu-bionic-ci-action@master
         with:
-          apt-dependencies: 'ruby-dev swig libeigen3-dev libignition-cmake2-dev'
+          apt-dependencies: |
```
mjcarroll, line 16: I would also propose moving apt-dependencies into it's own file.  It makes it easier to install all of the dependencies from multiple packages with a simple `find` command.  

I have a prototype here: https://github.com/ignitionrobotics/ign-bazel/blob/master/example/setup_workspace.sh#L9-L10

That may be out of the scope of this PR, but while we're breaking things...

chapulina: I like that, but it won't work for unreleased versions, right?

chapulina: Nevermind, I read that too quickly. I see you're using [this file](https://github.com/ignitionrobotics/ign-bazel/blob/master/.github/ci-bionic/packages.apt), I think that's good. Do you want to open a PR adding it to the action or should I?

chapulina: Fun story, I saw a comment of yours on a different PR and thought it was a reply to this telling me to open a PR. Then I went ahead and did it, just to realize what happened :grimacing: 

Well, here's the PR: https://github.com/ignition-tooling/ubuntu-bionic-ci-action/pull/15

PR #146, opened 2020-07-31
Round, Rounded, Correct, Distance(x, y, z, w) and operator< addition to Vector 4

### LINKED ISSUE #71: Add functions from Vector3 to Vector2 and Vector4

luccosta:
Resolves #71 .

This implements the functions `Round()`, `Rounded()`, `Correct()`, `Distance(T x, T y, T z, T w)` and operator overload `operator<(Vector4)` in `Vector4` API.

It enables:

```
vec4.Round();
Vector4 roundedVector = vec4.Rounded();
vec4.Correct();
vec4.Distance(1, 2, 3, 4);
if (vec4_a < vec4_b) {
  ...
}
```


luccosta:
I tried to test the `Correct` function generating an NaN first with `0/0` but this generate an Exception and them with math::sqrt(-1) that works fine, but generate the following message in code checker:

```
(warning) Passing value -1 to sqrt() leads to implementation-defined result.
```

chapulina:
> test the Correct function generating an NaN

While working on #139 I noticed that the code checker would complain about `0.0 / 0.0` but not `1.0 / 0.0`, maybe this is an option?

Also another option may be to use [std::nan()](https://en.cppreference.com/w/cpp/numeric/math/nan).

luccosta:
> but not 1.0 / 0.0, maybe this is an option?

Is a good option, I will check the result.

> Also another option may be to use std::nan().

Looks like a more consistent test, I will let with this.

luccosta:
> I will check the result.

With `1/0` I get `Floating-point exception` running the test as with `0/0`.

mjcarroll:
I don't think that you'll be able to get around the divide-by-zero issue either triggering the code check, compiler warning, or runtime error (since it's a signaling nan).

I agree that using `std::nan` here is the best approach, since that generates a quiet nan and won't trigger any floating point exceptions.

luccosta:
> I agree that using `std::nan` here is the best approach

Great, the current PR is using just `std::nan`.

PR #155, opened 2020-09-04
⬆️  6.5.0

chapulina:
Diff from 6.4.0: https://github.com/ignitionrobotics/ign-math/compare/ignition-math6_6.4.0...ign-math6

PR #157, opened 2020-09-04
Fix two windows warnings

JShep1:
Fixes two warnings found by @scpeters [here](https://github.com/ignitionrobotics/ign-math/pull/152#discussion_r483876964)

Unfortunately, I realized after the fact that I was passing in -1 into the `math::secNsecTimePoint()` function which takes type `const uint64_t` to indicate an invalid string. I suppose we're just going to have to return 0 if an invalid string is entered for now unless we want to add an extra parameter to `stringToTimePoint` that will be stored with the success of the operation.

Signed-off-by: John Shepherd <john@openrobotics.org>

### DIFF
--- a/include/ignition/math/Helpers.hh
+++ b/include/ignition/math/Helpers.hh
@@ -840,7 +840,7 @@ namespace ignition
         const std::string &_timeString)
     {
       std::chrono::steady_clock::time_point timePoint =
-        math::secNsecToTimePoint(-1, 0);
+        math::secNsecToTimePoint(0, 0);
 
       if (_timeString.empty())
         return timePoint;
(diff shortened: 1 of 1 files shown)


chapulina:
Oh just saw this, I also fixed the warnings here: https://github.com/ignitionrobotics/ign-math/pull/155/files

I think the tests will fail on this PR because the tests expect the negative time points.

JShep1:
> Oh just saw this, I also fixed the warnings here: https://github.com/ignitionrobotics/ign-math/pull/155/files
> 
> I think the tests will fail on this PR because the tests expect the negative time points.

Sounds good!  I don't think we'll honestly expect someone to enter time 2^64 -1 anyway, I'll close in favor of your solution

scpeters:
I just opened a PR to test this. I'll point it at #155 instead

PR #163, opened 2020-10-13
Add Capsule class with inertia calculation method

scpeters:
This package has classes for several shapes (`Box`, `Cylinder`, `Sphere`) as well as methods in `MassMatrix3` for computing inertial properties of these shapes assuming uniform density. This pull request adds a new `Capsule` shape class that is like a `Cylinder` with hemispheres capping each end. Methods for computing the inertial properties of a `Capsule` are added to `MassMatrix3` and validated in unit tests by comparing to the [formulae used by Open Dynamics Engine](https://bitbucket.org/odedevs/ode/src/0.16.2/ode/src/mass.cpp#lines-148:153).

Update: I've removed the `MassMatrix3` methods and removed the `Quaternion` parameter from the `Capsule` class.


include/ignition/math/MassMatrix3.hh
```diff
@@ -1106,6 +1106,107 @@ namespace ignition
...
+
+        // Inertia components of hemispheres
+        //  Centroid of hemisphere:
+        //    3/8 radius from flat base on axis of symmetry
+        //  Moment of inertia about centroid:
+        //    2/5 mass * radius^2 about axis of symmetry
+        //    83/320 mass * radius^2 about other axis
+        //    efunda.com/math/solids/solids_display.cfm?SolidName=Hemisphere
+        // add inertia of hemispheres about their centroid
+        L(0, 0) += sphereMass * 83. / 320. * radius2;
```
EricCousineau-TRI, line 1197: While these magic numbers are traceable, they seem a bit too magical, when really it's just the parallel-axis theorem with some solids (cylinder + hemisphere)?

Rather than bake this all together (which yeah, sure, it's great for efficiency, but how often will this be called?), why not leverage ign math to do parallel-axis with the constituent inertias?
Then you don't need comments, b/c the code is clear.

include/ignition/math/MassMatrix3.hh
```diff
@@ -1106,6 +1106,107 @@ namespace ignition
...
+      /// \param[in] _rot Rotational offset of equivalent capsule.
+      /// \return True if inertial properties were set successfully.
+      public: bool SetFromCapsuleZ(const Material &_mat,
+                                   const T _length,
+                                   const T _radius,
+                            const Quaternion<T> &_rot = Quaternion<T>::Identity)
+      {
+        // Check that density, _radius and _length are strictly positive
+        // and that quaternion is valid
+        if (_mat.Density() <= 0 || _length <= 0 || _radius <= 0 ||
```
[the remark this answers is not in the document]
EricCousineau-TRI: (Also, the fact that you have to make an overload named `SetFromCapsuleZ` means that now you have to redefine the frames that are already defined in the Capsule class itself?)

EricCousineau-TRI (approved):
Looks good at a high-level. The API here seems to be highly redundant and coupled (e.g. shapes + offsets, inertia computations, etc.), but understandably it'd be hard to refactor any time soon.

Made some picky comments, but nothing blocking.

Is there also a high-level issue that this is aiming towards?
e.g. https://github.com/osrf/sdformat/issues/176

EricCousineau-TRI:
Also, the shapes defined in `ign-math` seem like they have a shared contract, but there doesn't seem to be any form of contract enforcement (e.g. dynamic inheritance, static inheritance / mixins with CRTP, etc.).

Am I missing something there?

scpeters:
> Is there also a high-level issue that this is aiming towards?
> e.g. [osrf/sdformat#176](https://github.com/osrf/sdformat/issues/176) [osrf/sdformat#376](https://github.com/osrf/sdformat/issues/376)

it's in support of https://github.com/osrf/sdformat/issues/376, along with https://github.com/osrf/sdformat/pull/389

scpeters:
> The API here seems to be highly redundant and coupled (e.g. shapes + offsets, inertia computations, etc.), but understandably it'd be hard to refactor any time soon.

It took me a little while to remember, but I remembered a bit of the original intent behind this design. We added the `MassMatrix3` class (started in 2.3.0 and a bit more in 2.4.0) before adding the shape classes like `Sphere` and `Cylinder` (5.0.0). The `MassMatrix3` class is intended to represent a scalar mass and a 3x3 moment of inertia matrix. If you diagonalize the inertia matrix to get the principal moments of inertia and the orientation of its principal axes, then you can use these values along with its mass to compute the dimensions and orientation of a box of uniform density that has equivalent inertial properties (see [MassMatrix3::EquivalentBox](https://github.com/ignitionrobotics/ign-math/blob/ignition-math2_2.4.0/include/ignition/math/MassMatrix3.hh#L683-L727) from 2.4.0 for reference).

That `EquivalentBox` is used to visualize the moment of inertia in Gazebo. After adding `EquivalentBox`, we added its inverse `SetFromBox` as well as `SetFromSphere` and `SetFromCylinderZ` for good measure (we also use these in gazebo). The quaternion parameter was added to `SetFromBox` for symmetry with `EquivalentBox` and to `SetFromCylinderZ` to accommodate cylinders with different axis alignments.

When the shape functions were added in 5.0.0, they included a Quaternion parameter in the constructor, and in retrospect, I think they might be confusing things more than helping. I'll remove them from the `Capsule` constructors; they'll still be in the other shape classes, but we'll need a major version bump to get rid of those

I'll think further on whether we need `MassMatrix3::SetFromCapsuleZ` functions or not; if not I'll just put the math in the `Capsule` class

include/ignition/math/Capsule.hh
```diff
@@ -0,0 +1,189 @@
...
+    inline namespace IGNITION_MATH_VERSION_NAMESPACE {
+    //
+    /// \class Capsule Capsule.hh ignition/math/Capsule.hh
+    /// \brief A representation of a capsule or sphere-capped cylinder.
+    ///
+    /// The capsule class supports defining a capsule with a radius,
+    /// length, rotational offset, and material properties. The shape is
+    /// equivalent to a cylinder with capped with hemispheres. Radius and
+    /// length are in meters. See Material for more on material properties.
+    /// By default, a capsule's length is aligned with the Z axis. The
```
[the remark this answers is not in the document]
scpeters: I've removed the `Quaternion` from the `Capsule` class in 4de985ebddf5c104b6fce745a175d3c7a397d01e

do you think we still need to define frames in the documentation?

scpeters:
> I'll think further on whether we need `MassMatrix3::SetFromCapsuleZ` functions or not; if not I'll just put the math in the `Capsule` class

I thought about it, and we can use the `Inertial` class directly from `Capsule`, which we can't do from `MassMatrix3`, so I've done it in 0ac1488, and it's a bit more elegant

I've removed the `MassMatrix3::SetFromCapsuleZ` helper functions in d9c8d73

include/ignition/math/detail/Capsule.hh
```diff
@@ -0,0 +1,157 @@
...
+    this->material.SetDensity(newDensity);
+  return newDensity > 0;
+}
+
+//////////////////////////////////////////////////
+template<typename T>
+T Capsule<T>::DensityFromMass(const T _mass) const
+{
+  if (this->radius <= 0 || this->length <=0 || _mass <= 0)
+    return -1.0;
```
chapulina, line 150: I know this is following the pattern from other shapes, but I think we should consider returning `NaN` in such cases instead of `-1`.

include/ignition/math/detail/Capsule.hh
```diff
@@ -0,0 +1,157 @@
...
+  const T ixx = 83. / 320. * hemisphereMass * r2;
+  const T izz = 2. / 5.  * hemisphereMass * r2;
+  MassMatrix3<T> hemisphere(hemisphereMass, Vector3<T>(ixx, ixx, izz),
+                            Vector3<T>::Zero);;
+
+  // Distance from centroid of cylinder to centroid of hemisphere,
+  // since centroid of hemisphere is 3/8 radius from its flat base
+  const T dz = this->length / 2. + this->radius * 3. / 8.;
+  Inertial<T> upperCap(hemisphere, Pose3<T>(0, 0, dz, 0, 0, 0));
+  Inertial<T> lowerCap(hemisphere, Pose3<T>(0, 0, -dz, 0, 0, 0));
```
chapulina, line 117: I'm missing a 180 degree rotation here, but I guess it wouldn't make a difference since this is about the centroid?

chapulina (approved):
Looks good to me, I just have a question about the hemispheres, but I'm trusting the test that the math checks out.

include/ignition/math/detail/Capsule.hh
scpeters: yeah, it doesn't make a difference because it's about the centroid

include/ignition/math/detail/Capsule.hh
scpeters: what do you think @azeey?

include/ignition/math/Capsule.hh
```diff
@@ -0,0 +1,189 @@
...
+
+    // Inline bracket to help doxygen filtering.
+    inline namespace IGNITION_MATH_VERSION_NAMESPACE {
+    //
+    /// \class Capsule Capsule.hh ignition/math/Capsule.hh
+    /// \brief A representation of a capsule or sphere-capped cylinder.
+    ///
+    /// The capsule class supports defining a capsule with a radius,
+    /// length, rotational offset, and material properties. The shape is
+    /// equivalent to a cylinder with capped with hemispheres. Radius and
```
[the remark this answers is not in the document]
scpeters: this was fixed in 4de985ebddf5c104b6fce745a175d3c7a397d01e

include/ignition/math/MassMatrix3.hh
scpeters: I cleaned this up a bit in 0ac1488

include/ignition/math/MassMatrix3.hh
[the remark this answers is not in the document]
scpeters: I removed these duplicate functions in d9c8d73

include/ignition/math/Capsule.hh
```diff
@@ -0,0 +1,189 @@
...
+    /// length are in meters. See Material for more on material properties.
+    /// By default, a capsule's length is aligned with the Z axis. The
+    /// rotational offset encodes a rotation from the z axis.
+    template<typename Precision>
+    class Capsule
+    {
+      /// \brief Default constructor. The default radius and length are both
+      /// zero. The default rotational offset is
```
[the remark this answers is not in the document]
scpeters: I removed the rotational offset in 4de985e

scpeters:
> Also, the shapes defined in `ign-math` seem like they have a shared contract, but there doesn't seem to be any form of contract enforcement (e.g. dynamic inheritance, static inheritance / mixins with CRTP, etc.).
> 
> Am I missing something there?

no I don't think you're missing anything. I think the shape objects have similar API's but it's not enforced

include/ignition/math/Capsule.hh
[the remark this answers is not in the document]
EricCousineau-TRI: Nah, if there's just one frame, it's a lot easier to reason about locally. Thanks!

EricCousineau-TRI:
Aye, sounds good. Minor concern is that the `Capsule` API may now deviates from others (given that it computes its own inertia, rather than kinda doing some psuedo static-double-dispatch things).

include/ignition/math/detail/Capsule.hh
azeey: I'm not confident if returning a `NaN` is good practice, but the math functions in the standard c++ library either throw an exception or return a `NaN` when the input is invalid, so I think it's reasonable to return a `NaN`.

include/ignition/math/Capsule.hh
```diff
@@ -0,0 +1,159 @@
...
+    /// The capsule class supports defining a capsule with a radius,
+    /// length, and material properties. The shape is equivalent to a cylinder
+    /// aligned with the Z-axis and capped with hemispheres. Radius and
+    /// length are in meters. See Material for more on material properties.
+    template<typename Precision>
+    class Capsule
+    {
+      /// \brief Default constructor. The default radius and length are both
+      /// zero.
+      public: Capsule() = default;
```
azeey, line 45: What do you think about not providing a default constructor? That way the only time we would have invalid Capsule objects would be if the input length and radius values are invalid.

src/Capsule_TEST.cc
```diff
@@ -0,0 +1,127 @@
...
+
+  const double cylinderVolume = IGN_PI * r*r * l;
+  const double sphereVolume = IGN_PI * 4. / 3. * r*r*r;
+  const double volume = cylinderVolume + sphereVolume;
+  const double cylinderMass = mass * cylinderVolume / volume;
+  const double sphereMass = mass * sphereVolume / volume;
+
+  // expected values based on formula used in Open Dynamics Engine
+  // https://bitbucket.org/odedevs/ode/src/0.16.2/ode/src/mass.cpp#lines-148:153
+  double ixxIyy = (1/12.0) * cylinderMass * (3*r*r + l*l)
```
azeey, line 114: I checked this against DART (https://github.com/dartsim/dart/blob/master/dart/dynamics/CapsuleShape.cpp#L123) and there is a discrepancy, but I think what you have here is correct. The referenced page (https://www.gamedev.net/tutorials/_/technical/math-and-physics/capsule-inertia-tensor-r3856/) aligns with the formula here except the `0.25*l*l` term here vs `0.5*l*l` term there. Scrolling down further, I see a comment that it should actually be `0.25*l*l`.

scpeters: it shows `H^2 m / 4` in equation 12 on that page; I think there was just a transcription error in equation 14

scpeters: I picked ODE because I think it's right, though it's not authoritative and doesn't have any justification. I could add a link to that gamedev.net page as well if you think that's a good reference, despite its inconsistency

include/ignition/math/Capsule.hh
```diff
@@ -0,0 +1,159 @@
...
+      /// \return The length of the capsule in meters.
+      public: Precision Length() const;
+
+      /// \brief Set the length in meters.
+      /// \param[in] _length The length of the capsule in meters.
+      public: void SetLength(const Precision _length);
+
+      /// \brief Set the length in meters.
+      /// \param[in] _length The length of the capsule in meters.
+      public: void SetLength(const Precision _length) const;
```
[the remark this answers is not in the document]
scpeters: yes, I will remove this

include/ignition/math/Capsule.hh
```diff
@@ -0,0 +1,159 @@
...
+    /// length are in meters. See Material for more on material properties.
+    template<typename Precision>
+    class Capsule
+    {
+      /// \brief Default constructor. The default radius and length are both
+      /// zero.
+      public: Capsule() = default;
+
+      /// \brief Construct a capsule with a length, radius, and optionally
+      /// a rotational offset.
```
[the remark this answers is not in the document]
scpeters: nope! I'll remove it

include/ignition/math/Capsule.hh
```diff
@@ -0,0 +1,159 @@
...
+    inline namespace IGNITION_MATH_VERSION_NAMESPACE {
+    //
+    /// \class Capsule Capsule.hh ignition/math/Capsule.hh
+    /// \brief A representation of a capsule or sphere-capped cylinder.
+    ///
+    /// The capsule class supports defining a capsule with a radius,
+    /// length, and material properties. The shape is equivalent to a cylinder
+    /// aligned with the Z-axis and capped with hemispheres. Radius and
+    /// length are in meters. See Material for more on material properties.
+    template<typename Precision>
```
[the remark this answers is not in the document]
scpeters: added in f0419f19dde94831cbe2aed4c4599e3cbb68e65b

include/ignition/math/Capsule.hh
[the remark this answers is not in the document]
scpeters: f0419f19dde94831cbe2aed4c4599e3cbb68e65b

include/ignition/math/Capsule.hh
```diff
@@ -0,0 +1,159 @@
...
+
+      /// \brief Construct a capsule with a length, radius, and material.
+      /// \param[in] _length Length of the capsule.
+      /// \param[in] _radius Radius of the capsule.
+      /// \param[in] _mat Material property for the capsule.
+      public: Capsule(const Precision _length, const Precision _radius,
+                  const Material &_mat);
+
+      /// \brief Destructor
+      public: ~Capsule() = default;
```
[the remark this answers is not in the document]
scpeters: sure, f0419f19dde94831cbe2aed4c4599e3cbb68e65b

include/ignition/math/Capsule.hh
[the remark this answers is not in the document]
scpeters: f0419f19dde94831cbe2aed4c4599e3cbb68e65b

include/ignition/math/detail/Capsule.hh
scpeters: we don't use `NaN`s much in our APIs, and I'm careful about introducing them because they can quickly spread and it can be hard to figure out where they came from. In this case, I think it's fine because this function isn't widely used

scpeters: 2628fabf38785957845ac985d29ea0b3cce07613

include/ignition/math/detail/Capsule.hh
```diff
@@ -0,0 +1,157 @@
...
+
+#include <ignition/math/Inertial.hh>
+
+namespace ignition
+{
+namespace math
+{
+
+//////////////////////////////////////////////////
+template<typename T>
```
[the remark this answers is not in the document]
scpeters: the other shape classes are using `T` in their `detail` headers. I'll leave it as `T` for now since it's so much easier to read; we can revisit if there's a more concrete example of the problem

include/ignition/math/Capsule.hh
scpeters: I like having the default constructor. If you want to create a shape without remembering the order of the arguments in the constructor:

~~~
Capsuled capsule;
capsule.SetLength(length);
capsule.SetRadius(radius);
~~~

if other people don't like it, we can remove it, but I'm inclined to keep it

include/ignition/math/Capsule.hh
```diff
@@ -0,0 +1,159 @@
...
+
+      /// \brief Get the mass matrix for this capsule. This function
+      /// is only meaningful if the capsule's radius, length, and material
+      /// have been set.
+      /// \param[out] _massMat The computed mass matrix will be stored
+      /// here.
+      /// \return False if computation of the mass matrix failed, which
+      /// could be due to an invalid radius (<=0), length (<=0), or density
+      /// (<=0).
+      public: bool MassMatrix(MassMatrix3<Precision> &_massMat) const;
```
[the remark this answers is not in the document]
scpeters: that's a good idea; it makes clear that invalid `MassMatrix3` values won't be returned. I've made the change in d379a5b366422e3b86fdac865e6d8a335031a11c

src/Capsule_TEST.cc
azeey: The ODE reference is fine. I just checked DART to verify and saw it was different.

include/ignition/math/detail/Capsule.hh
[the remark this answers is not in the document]
azeey: Sounds good.

azeey (approved):

include/ignition/math/Capsule.hh
mxgrey: If I put on my pedantry hat, the "right" way to do what @scpeters is suggesting would be to have something like a `Capsule::Builder` class that lets you fill in the information necessary to build a `Capsule` object, and then you could call `build()` on it (or something similar) and it constructs a capsule for you, perhaps throwing an exception if the parameters are invalid (e.g. radius <= 0, length <= 0).

Some would (fairly) argue that's overkill for this use case, but the more I've had to deal with writing production-quality code, the more I've been feeling it's worth the effort to make every assumption explicit inside the code itself.

Issue #34, opened 2015-12-01
Vector3::Normalize returns Vector3, Vector[24]::Normalize return void


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-math/issues/34)) by Steve Peters (Bitbucket: [Steven Peters](https://bitbucket.org/%7B2ccfed09-18b8-4921-8d58-15ef01092802%7D/), GitHub: [scpeters](https://github.com/scpeters)).**

----------------------------------------
We should make these consistent.

pxalcantara:
> I think we could do a similar thing for Normalize with the vector classes and perhaps Quaternion as well:

I think this bug is fixed in the `Vector` class. About the `Quaternion` to solve this issue, we should add **Normalized** to the class?

chapulina:
I think this is still inconsistent:

https://github.com/ignitionrobotics/ign-math/blob/b5306cc9ea8e7ac2075547b227eb5e6c738efb3e/include/ignition/math/Vector3.hh#L133

https://github.com/ignitionrobotics/ign-math/blob/b5306cc9ea8e7ac2075547b227eb5e6c738efb3e/include/ignition/math/Vector2.hh#L102

https://github.com/ignitionrobotics/ign-math/blob/b5306cc9ea8e7ac2075547b227eb5e6c738efb3e/include/ignition/math/Vector4.hh#L149

I agree with the solution proposed by @scpeters above, the problem is that it will be too disruptive. We can't tick-tock deprecate `Vector3::Normalize()`, we would have to change the return type from `Vector3` to `void` on `ign-math7`. I think we need to balance the API improvement with the impact on users. Maybe since a lot will change in `ign-math7` due to #101, maybe it's a good time to also make this change? What do you think, @scpeters ?

> About the Quaternion to solve this issue, we should add Normalized to the class?

Nice catch, it already has `void Normalize`, so we could add `Quaternion<T> Normalized`.

https://github.com/ignitionrobotics/ign-math/blob/b5306cc9ea8e7ac2075547b227eb5e6c738efb3e/include/ignition/math/Quaternion.hh#L280-L281

pxalcantara:
> Nice catch, it already has void Normalize, so we could add Quaternion<T> Normalized.

So, I'm planning to work on the implementation of the `Quaternion<T> Normalized` method.

PR #169, opened 2020-10-20
Add Normalized method to Quaternion.hh

pxalcantara:
Add the method Normalized to the Quaternion class.

This PR is related to #34

### DIFF
--- a/src/Quaternion_TEST.cc
+++ b/src/Quaternion_TEST.cc
@@ -252,6 +252,8 @@ TEST(QuaternionTest, Math)
   q.Normalize();
   EXPECT_TRUE(q == math::Quaterniond(0.182574, 0.365148, 0.547723, 0.730297));
 
+  math::Quaterniond q2 = q.Normalized();
+  EXPECT_TRUE(q2 == math::Quaterniond(0.182574, 0.365148, 0.547723, 0.730297));
 
   EXPECT_TRUE(math::equal(q.Roll(), 1.4289, 1e-3));
   EXPECT_TRUE(math::equal(q.Pitch(), -0.339837, 1e-3));
--- a/src/Quaternion_TEST.cc
+++ b/src/Quaternion_TEST.cc
@@ -249,11 +249,12 @@ TEST(QuaternionTest, Math)
   EXPECT_TRUE(math::equal(q.Y(), 3.0));
   EXPECT_TRUE(math::equal(q.Z(), 4.0));
 
+  math::Quaterniond q2 = q.Normalized();
+  EXPECT_TRUE(q2 == math::Quaterniond(0.182574, 0.365148, 0.547723, 0.730297));
+
   q.Normalize();
   EXPECT_TRUE(q == math::Quaterniond(0.182574, 0.365148, 0.547723, 0.730297));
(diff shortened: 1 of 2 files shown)


pxalcantara:
@chapulina @j-rivero Could you help me with this CI error for Windows? I couldn't get the cause :thinking: 

By the way, Why the unit tests case for Quaternion.hh is so [big](https://github.com/ignitionrobotics/ign-math/blob/2981c125c458d7dac097fc6e014901e72d691fac/src/Quaternion_TEST.cc#L215)? Shouldn't we split it into several test cases for each method, in order to be easier to detect the error?

src/Quaternion_TEST.cc
```diff
@@ -252,6 +252,8 @@ TEST(QuaternionTest, Math)
   q.Normalize();
   EXPECT_TRUE(q == math::Quaterniond(0.182574, 0.365148, 0.547723, 0.730297));
 
+  math::Quaterniond q2 = q.Normalized();
```
chapulina, line 255: Isn't `q` already normalized in this line? So in this case `q2 == q`, right?

How about moving this test case above line 252, so that the vector returned in the `Normalized` call is different from the original vector?

chapulina (commented):
> Why the unit tests case for Quaternion.hh is so big? Shouldn't we split it into several test cases for each method, in order to be easier to detect the error?

I think that's probably just old. I agree that having smaller tests would make it easier to handle regressions whenever we have any.

chapulina:
I couldn't tell what was wrong with the Windows build, so I retriggered it

j-rivero:
> I couldn't tell what was wrong with the Windows build, so I retriggered it

I see an "internal compiler error". Usually means one of two things:
 * There is not enough RAM to compile the code 
 * The code change triggered a bug in the compiler
 
We have run 869 and 870 on the same Windows node with this problem. I've changed the configuration to run on the other Windows node, let's see if it fails too: [Build Status](https://build.osrfoundation.org/job/ignition_math-ci-pr_any-windows7-amd64/871/)

src/Quaternion_TEST.cc
pxalcantara: Yes, sure, I'll do it!

pxalcantara:
> > Why the unit tests case for Quaternion.hh is so big? Shouldn't we split it into several test cases for each method, in order to be easier to detect the error?
> 
> I think that's probably just old. I agree that having smaller tests would make it easier to handle regressions whenever we have any.

Can I make these changes? Should I open an issue for that?

chapulina:
> Can I make these changes?

Yeah go for it! Thanks!

pxalcantara:
> > Can I make these changes?
> 
> Yeah go for it! Thanks!

Just make PR but don't need to open a new issue right?

chapulina:
> Just make PR but don't need to open a new issue right?

Yeah I think that's fine :ok:

src/Quaternion_TEST.cc
pxalcantara: @chapulina Done! :wink:

chapulina (approved):

Issue #60, opened 2017-02-02
Pose addition and subtraction needs work [gazebo issue 216]


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-math/issues/60)) by Louise Poubel (Bitbucket: [chapulina](https://bitbucket.org/%7B5cfa2075-477b-4ded-bdb9-8d2479544ec4%7D/), GitHub: [chapulina](https://github.com/chapulina)).**

----------------------------------------
Moving Gazebo issue 216 here now that gazebo::math has been fully deprecated. Please consider the discussion on that issue:

https://bitbucket.org/osrf/gazebo/issues/216/pose-addition-and-subtraction-needs-work

traversaro:
I think this has been fixed (unfortunately with breaking change in the behavior of `operator*(Pose3d, Pose3d)` in ignition-math6 with PR https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-math/pull-requests/301 right? See https://github.com/ignitionrobotics/ign-math/blob/master/Changelog.md#ignition-math-620 .

chapulina:
I think there's still the question of whether we want to remove the addition and subtraction operators:

https://github.com/ignitionrobotics/ign-math/blob/6a3c4a32b18e6bd7956fecfe53ccfe14bfcd2907/include/ignition/math/Pose3.hh#L183-L189

scpeters:
yeah, I would support deprecating them to encourage the use of `*`

traversaro:
> yeah, I would support deprecating them to encourage the use of `*`

I personally agree. `+` is often used to refer to commutative operations, and using it for a non-commutative operation is quite confusing.

scpeters:
I'll open a pull request to deprecate the `+` and `-` operators for ign-math7 and see how much trouble it causes

scpeters:
Pose `+` and `-` operators are now deprecated in `math7`

PR #170, opened 2020-10-23
Document Pose::operator*()

sloretz:
Adds some documentation for `Pose::operator*()`. If I understand correctly, it looks like `B + A` is the same as `A * B`, but [operator+ is discouraged](https://github.com/ignitionrobotics/ign-math/issues/60).

### DIFF
--- a/include/ignition/math/Pose3.hh
+++ b/include/ignition/math/Pose3.hh
@@ -258,6 +258,9 @@ namespace ignition
       }
 
       /// \brief Multiplication operator.
+      /// A is the transform from O to P specified in frame O
+      /// B is the transform from P to Q specified in frame P
+      /// then, A * B is the transform from O to Q specified in frame O
       /// \param[in] _pose The pose to mutlipy by.
       /// \return The resulting pose.
       public: Pose3<T> operator*(const Pose3<T> &_pose) const
--- a/include/ignition/math/Pose3.hh
+++ b/include/ignition/math/Pose3.hh
@@ -258,9 +258,8 @@ namespace ignition
       }
 
       /// \brief Multiplication operator.
-      /// A is the transform from O to P specified in frame O
-      /// B is the transform from P to Q specified in frame P
-      /// then, A * B is the transform from O to Q specified in frame O
+      /// Given X_OP (frame P relative to O) and X_PQ (frame Q relative to P)
+      /// then X_OQ = X_OP * X_PQ (frame Q relative to O).
       /// \param[in] _pose The pose to mutlipy by.
       /// \return The resulting pose.
       public: Pose3<T> operator*(const Pose3<T> &_pose) const


chapulina:
How about adding a couple of illustrative test cases? I don't see that operator being explicitly tested here:

https://github.com/ignitionrobotics/ign-math/blob/main/src/Pose_TEST.cc

Maybe it's tested indirectly from somewhere else.

scpeters:
we added tests for `Pose*` in `Matrix4_TEST` when we changed its behavior to resolve #60:

* https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-math/pull-requests/301/page/1
* https://github.com/ignitionrobotics/ign-math/compare/9552cd5ce7eaed6651b726fc03a23b637db83675..7b51f5610930caa03966cf41bf2014bf547cd749#diff-264517f0fd7cdc07fe41f4522b6a18fb832db7215be636d33eae8cb7084ae8e4

scpeters:
for the life of me, I have the toughest time verbalizing how these transforms work; I'm sure @azeey remembers how much trouble I had when we were writing the following SDFormat tutorial:

* http://sdformat.org/tutorials?tut=specify_pose&cat=specification&

I'll let him speak up if he has any concrete suggestions for syntax

azeey:
Thanks for adding the documentation @sloretz. In most robotics books (eg. http://hades.mech.northwestern.edu/images/2/25/MR-v2.pdf), the `*` operator is used for composing transforms rather than `+`. The notation given in these books uses two frames to describe a transform/pose. In http://sdformat.org/tutorials?tut=specify_pose&cat=specification&, we have used `X_BA` to indicate the transform from B to A. This is also the same quantity as the pose of frame A (relative to and expressed in) frame B. 

Given two `Pose` objects, the order of arguments when using `Pose::operator*` is the same as if the `Pose`s were first converted to 4x4 homogeneous matrices and the product of the matrices was taken. So if we have frames `O`, `P`, and `Q` and we have the poses `X_OP` (frame `P` relative to `O`) and `X_PQ` (frame `Q` relative to `P`), then `X_OQ` (frame `Q` relative to `O`) would be `X_OQ = X_OP * X_PQ`

sloretz:
> Given two Pose objects, the order of arguments when using Pose::operator* is the same as if the Poses were first converted to 4x4 homogeneous matrices and the product of the matrices was taken. So if we have frames O, P, and Q and we have the poses X_OP (frame P relative to O) and X_PQ (frame Q relative to P), then X_OQ (frame Q relative to O) would be X_OQ = X_OP * X_PQ

Uses this notation in 975a50c

azeey (approved):
LGTM

sloretz:
Does this need a second reviewer, or is it good to merge? If the latter, would someone mind merging it? I don't have write access to the ignitionrobotics repos.

chapulina (approved):
:ok_hand:

PR #171, opened 2020-10-23
Bazel Updates for math6

mjcarroll:
Rebase the bazel branch to be off of math6, the version used in dome.

Specific cleanups and feedback incorporated in 18b35a0

This will replace the current `bazel` branch.


eigen3/BUILD.bazel
```diff
@@ -0,0 +1,45 @@
...
+package(
+    default_visibility = IGNITION_VISIBILITY,
+    features = [
+        "-layering_check",
+    ],
+)
+
+licenses(["notice"])
+
+exports_files(["LICENSE"])
```
mahiuchun, line 16: nit: the convention we use is that licenses in each BUILD file while exports_files only in the top-level BUILD file.

eigen3/BUILD.bazel
```diff
@@ -0,0 +1,45 @@
+load(
+    "//ign_bazel:build_defs.bzl",
+    "IGNITION_ROOT",
+    "IGNITION_VISIBILITY",
+)
+
+package(
+    default_visibility = IGNITION_VISIBILITY,
+    features = [
+        "-layering_check",
```
mahiuchun, line 10: nit: `-layering_check` is not necessary in this case.

JShep1 (approved):
Builds and runs fine with all tests passing for `bazel` and `colcon`

.github/ci-bionic/packages.apt
```diff
@@ -0,0 +1,3 @@
+ruby-dev
+swig
+libeigen3-dev
```
chapulina, line 3: These are all on `.github/ci/packages.apt` already

https://github.com/ignitionrobotics/ign-math/blob/ign-math6/.github/ci/packages.apt

eigen3/BUILD.bazel
mjcarroll: Removed in e7a37b4

eigen3/BUILD.bazel
mjcarroll: Removed  in e7a37b4

scpeters:
are we going to setup CI for bazel to prevent regressions?

JShep1:
Just added the `Capsule` class.  This brings up a good point, perhaps we eventually want to glob for files?  We're globbing for tests so it will automatically break CI if someone adds new files and tests for them since the bazel build files will pick up the new tests but not the new files.

chapulina (approved):
I don't know much about Bazel, but I didn't notice anything weird, and CI that's using this branch seems to be happy with `ign-math`:

https://github.com/ignitionrobotics/ign-bazel/runs/2068842483?check_suite_focus=true

The branch for that build is configured here:

https://github.com/ignitionrobotics/ign-bazel/blob/6c7128cb4833bbbb06d397f785c61d02d7f8b962/.github/ci/bazel.repos#L5

PR #172, opened 2020-10-26
Add construction tests for Vector2, Vector3, Vector4, Quaternion, and Pose

### LINKED ISSUE #76: Many objects are not std::move-able
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-math/issues/76)) by Steve Peters (Bitbucket: [Steven Peters](https://bitbucket.org/%7B2ccfed09-18b8-4921-8d58-15ef01092802%7D/), GitHub: [scpeters](https://github.com/scpeters)).**

In order to use `std::move` semantics in c++ (like assignment of unique_ptr's), objects need to have move constructor and move assignment operators defined (also known as the [rule of 5](https://en.wikipedia.org/wiki/Rule_of_three_(C%2B%2B_programming)#Rule_of_Five). This lets you do things like this:

~~~
std::unique_ptr<math::Pose3d> ptr1, ptr2;
ptr1.reset(new math::Pose3d);
ptr2 = ptr1;
~~~

but that doesn't currently work because we aren't following the rule of 5 for many of our classes. I mention this because I was trying to use it in a gazebo plugin and it didn't work. I've added some statements to unit tests in b25df8976d840625395e66cd8a0966d69d1405f6 that illustrate this failure to compile.

nlamprian:
The classes are already moveable. The error reported in the issue comes from the fact that the operation is a copy assignment to unique_ptr, and not a move assignment.

Closes #76

Signed-off-by: Nick Lamprianidis <nlamprian@gmail.com>

### DIFF
--- a/src/Vector4_TEST.cc
+++ b/src/Vector4_TEST.cc
@@ -24,6 +24,38 @@
 using namespace ignition;
 
 /////////////////////////////////////////////////
+TEST(Vector4dTest, Construction)
+{
+  math::Vector4d vec(1, 0, 0, 0);
+
+  // Copy constructor
+  math::Vector4d vec2(vec);
+  EXPECT_EQ(vec2, vec);
+
+  // Copy operator
+  math::Vector4d vec3;
+  vec3 = vec;
+  EXPECT_EQ(vec3, vec);
+
+  // Move constructor
+  math::Vector4d vec4(std::move(vec));
+  EXPECT_EQ(vec4, vec2);
+  vec = vec4;
+  EXPECT_EQ(vec, vec2);
+
+  // Move operator
+  math::Vector4d vec5;
+  vec5 = std::move(vec2);
+  EXPECT_EQ(vec5, vec3);
+  vec2 = vec5;
+  EXPECT_EQ(vec2, vec3);
+
+  // Inequality
+  math::Vector4d vec6;
+  EXPECT_NE(vec6, vec3);
+}
+
+/////////////////////////////////////////////////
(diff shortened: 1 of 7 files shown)


ahcorde (approved):

src/Vector4_TEST.cc
```diff
@@ -23,6 +23,38 @@
 
 using namespace ignition;
 
+/////////////////////////////////////////////////
+TEST(Vector4dTest, Construction)
+{
+    math::Vector4d vec(1, 0, 0, 0);
```
chapulina, line 29: Nitpick: use 2 spaces indentation

chapulina (approved):
> The error reported in the issue comes from the fact that the operation is a copy assignment to unique_ptr, and not a move assignment.

Thanks for checking and for the tests. You're right, I checked that the example tests work by changing 

`ptr2 = ptr1;`

to 

`ptr2 = std::move(ptr1);`

src/Vector4_TEST.cc
nlamprian: fixed

nlamprian:
Should these tests be replicated for the rest of the classes?

chapulina:
> Should these tests be replicated for the rest of the classes?

That would be great if you're up for it. Hopefully you won't find any classes that aren't movable. And having the tests in place will make sure we don't break this usage in the future.

But when it comes to #76, I think this is enough to address those concerns, so we can close it.

PR #178, opened 2020-11-13
6 to main

adlarkin:
Forward port `ign-math6` to `main`.

### DIFF
--- a/src/Quaternion_TEST.cc
+++ b/src/Quaternion_TEST.cc
@@ -27,6 +27,38 @@
 using namespace ignition;
 
 /////////////////////////////////////////////////
+TEST(QuaternionTest, Construction)
+{
+  math::Quaterniond q(0, 0, 0, 1);
+
+  // Copy constructor
+  math::Quaterniond q2(q);
+  EXPECT_EQ(q2, q);
+
+  // Copy operator
+  math::Quaterniond q3;
+  q3 = q;
+  EXPECT_EQ(q3, q);
+
+  // Move constructor
+  math::Quaterniond q4(std::move(q));
+  EXPECT_EQ(q4, q2);
+  q = q4;
+  EXPECT_EQ(q, q2);
+
+  // Move operator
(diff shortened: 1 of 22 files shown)


src/Quaternion_TEST.cc
```diff
@@ -235,46 +267,93 @@ TEST(QuaternionTest, Integrate)
...
+            math::Quaterniond(0.545456, -0.588972, 0.093284, 0.588972));
 
   math::Quaterniond q1 = q;
-  q1.SetW(2.0);
-  EXPECT_TRUE(q1.Log() ==
-      math::Quaterniond(0, -0.698401, 0.110616, 0.698401));
+  q1.X(0.000000001);
+  q1.Y(0.0);
+  q1.Z(0.0);
+  q1.W(0.0);
```
chapulina, line 293: These tests should use the new `Set*` APIs, because these functions are deprecated. See the deprecation warnings:

https://build.osrfoundation.org/job/ignition_math-ci-pr_any-homebrew-amd64/743/clang/new/

chapulina (changes requested):

src/Quaternion_TEST.cc
adlarkin: Thanks for catching this - I believe I updated all deprecated function calls in https://github.com/ignitionrobotics/ign-math/pull/178/commits/02530ea4a9d4789603d75f9215312dc051f1223c.

Just to make sure I understand ignition deprecations correctly, does this mean that this function will become deprecated in `ign-math7`?

https://github.com/ignitionrobotics/ign-math/blob/560b43e314f92131f0113fd150853e89c471a0f2/include/ignition/math/Quaternion.hh#L310

chapulina: Correct, `IGN_DEPRECATED` gives a compile time warning whenever this function is used. This will happen for users of `ign-math7`. Then for `ign-math8`, this function will be removed.

chapulina (approved):

PR #179, opened 2020-11-19
Prepare for 6.7.0 release

scpeters:
Changelog and version bump.

Diff from 6.6.0: https://github.com/ignitionrobotics/ign-math/compare/ignition-math6_6.6.0...ign-math6


chapulina (approved):
The only PR that I think could be added to the changelog is #162.

scpeters:
> The only PR that I think could be added to the changelog is #162.

good call: ffc89bcc63953dd80aa3e417d8d47bfd5fddd5d2

PR #182, opened 2020-12-11
Include ellipsoid as a math shape

brawner:
This follows #163 pretty closely, but I'm happy to make changes where needed for this own class.

I noticed that the API for different shapes has diverged, and it may also be worth revisiting how bad parameters are dealt with across the different shapes.

Signed-off-by: Stephen Brawner <brawner@gmail.com>

### DIFF
--- a/include/ignition/math/Ellipsoid.hh
+++ b/include/ignition/math/Ellipsoid.hh
@@ -25,9 +25,6 @@ namespace ignition
 {
   namespace math
   {
-    // Foward declarations
-    class EllipsoidPrivate;
-
     // Inline bracket to help doxygen filtering.
     inline namespace IGNITION_MATH_VERSION_NAMESPACE {
     //
--- a/include/ignition/math/Ellipsoid.hh
+++ b/include/ignition/math/Ellipsoid.hh
@@ -47,7 +47,7 @@ namespace ignition
 
       /// \brief Construct a ellipsoid with a Vector3 of three radii.
       /// \param[in] _radii The three radii (x, y, z) defining this ellipsoid
-      public: Ellipsoid(const Vector3<Precision> &_radii);
+      public: explicit Ellipsoid(const Vector3<Precision> &_radii);
 
       /// \brief Construct a ellipsoid with three radii and a material.
       /// \param[in] _radii The three radii (x, y, z) defining this ellipsoid
--- a/include/ignition/math/Ellipsoid.hh
+++ b/include/ignition/math/Ellipsoid.hh
@@ -14,8 +14,8 @@
  * limitations under the License.
  *
 */
-#ifndef IGNITION_MATH_CAPSULE_HH_
-#define IGNITION_MATH_CAPSULE_HH_
+#ifndef IGNITION_MATH_ELLIPSOID_HH_
+#define IGNITION_MATH_ELLIPSOID_HH_
 
 #include <optional>
 #include "ignition/math/MassMatrix3.hh"
--- a/include/ignition/math/Ellipsoid.hh
+++ b/include/ignition/math/Ellipsoid.hh
@@ -1,0 +1,137 @@
+/*
+ * Copyright (C) 2020 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+#ifndef IGNITION_MATH_CAPSULE_HH_
+#define IGNITION_MATH_CAPSULE_HH_
+
+#include <optional>
+#include "ignition/math/MassMatrix3.hh"
+#include "ignition/math/Material.hh"
+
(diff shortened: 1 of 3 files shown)


include/ignition/math/Ellipsoid.hh
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+#ifndef IGNITION_MATH_CAPSULE_HH_
+#define IGNITION_MATH_CAPSULE_HH_
```
scpeters, line 18: replace `CAPSULE` with `ELLIPSOID` in the header guards

include/ignition/math/detail/Ellipsoid.hh
```diff
@@ -0,0 +1,128 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+#ifndef IGNITION_MATH_DETAIL_CAPSULE_HH_
+#define IGNITION_MATH_DETAIL_CAPSULE_HH_
```
scpeters, line 18: replace `CAPSULE` with `ELLIPSOID` in the header guards

include/ignition/math/detail/Ellipsoid.hh
```diff
@@ -0,0 +1,128 @@
...
+{
+  return this->radii == _ellipsoid.radii &&
+    this->material == _ellipsoid.material;
+}
+
+//////////////////////////////////////////////////
+template<typename T>
+std::optional< MassMatrix3<T> > Ellipsoid<T>::MassMatrix() const
+{
+  if (this->radii.X() < 0 || this->radii.Y() < 0 || this->radii.Z() < 0)
```
scpeters, line 79: I think we want to use `<=` for these comparisons, since the inertial properties are invalid if any radii are `0`

scpeters (commented):
looks straightforward; I just have a few minor comments

include/ignition/math/Ellipsoid.hh
brawner: Fixed here and below

include/ignition/math/detail/Ellipsoid.hh
brawner: Done

include/ignition/math/detail/Ellipsoid.hh
brawner: Completely reasonable and fixed, but then that begs another question. If 0 and negative values are invalid, should they be checked during construction/`SetRadii()`? And an exception thrown that case?

brawner:
Forgot to update a unit test for the stricter check that was added.

include/ignition/math/detail/Ellipsoid.hh
scpeters: > If 0 and negative values are invalid, should they be checked during construction/`SetRadii()`? And an exception thrown in that case?

Currently the constructors and Get/Set methods act like plain old data structures, and there's only validity checking in the methods dealing with inertia computations. I'm inclined to keep it that way for `ign-math6` so that the shape classes have the same exception behavior. I think we could consider that for `ignition-math7`, but it would be good to consider the impact it would have on downstream code in libsdformat, ign-gazebo, etc.

include/ignition/math/detail/Ellipsoid.hh
```diff
@@ -0,0 +1,128 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+#ifndef IGNITION_MATH_DETAIL_ELLIPSE_HH_
+#define IGNITION_MATH_DETAIL_ELLIPSE_HH_
```
scpeters, line 18: nit: `ELLIPSE` -> `ELLIPSOID`

brawner: Fixed

brawner:
@scpeters Is there anything else you would like to see from this PR?

include/ignition/math/Ellipsoid.hh
```diff
@@ -0,0 +1,137 @@
...
+#include <optional>
+#include "ignition/math/MassMatrix3.hh"
+#include "ignition/math/Material.hh"
+
+namespace ignition
+{
+  namespace math
+  {
+    // Foward declarations
+    class EllipsoidPrivate;
```
chapulina, line 29: Did you mean to create this class? I see the same thing sneaked into the `Capsule` class.

scpeters: I'll take the liberty of removing this so it can rerun the windows job

scpeters: there is no `CapsulePrivate`, so that forward declaration could be removed as well

scpeters (approved):

include/ignition/math/Ellipsoid.hh
brawner: Thanks @scpeters. We can blame my gratuitous use of copy/paste.

brawner:
Thanks again @scpeters for the review and quick fix. Looks like all checks now pass.

PR #183, opened 2020-12-15
Avoid moving a return value, it might prevent (N)RVO

ivanpauno:
I'm building in Focal, this PR fixes some "-Wno-pessimizing-move" warnings (which does not happen with the gcc version in Bionic).

From [here](https://gcc.gnu.org/onlinedocs/gcc-9.1.0/gcc/C_002b_002b-Dialect-Options.html):

```
-Wno-pessimizing-move (C++ and Objective-C++ only)
This warning warns when a call to std::move prevents copy elision. A typical scenario when copy elision can occur is when returning in a function with a class return type, when the expression being returned is the name of a non-volatile automatic object, and is not a function parameter, and has the same type as the function return type.

struct T {
…
};
T fn()
{
  T t;
  …
  return std::move (t);
}
But in this example, the std::move call prevents copy elision.

This warning is enabled by -Wall.
```


chapulina (approved):
Thanks for catching that. We should enable more warnings on our Jenkins CI builds. Or maybe they'll already be enabled once we move CI to Focal.

ivanpauno:
> We should enable more warnings on our Jenkins CI builds. Or maybe they'll already be enabled once we move CI to Focal.

It may be a combination of both things, the focal build already has some warnings in the logs that aren't being captured + current focal CI is using GCC 8 and I'm using the default Ubuntu Focal compiler (gcc 9.3.0).

---

The `ignition_math-ci-pr_any-homebrew-amd64` failure seems unrelated to me.

PR #194, opened 2021-03-04
Add speed limiter class

chapulina:

# 🎉 New feature

## Summary

We've been using the `SpeedLimiter` class ported from [`ros_control`](https://github.com/ros-controls/ros_controllers/blob/melodic-devel/diff_drive_controller/src/speed_limiter.cpp) on Ignition Gazebo's diff drive controller for a while. Now another controller is being added and wants to use the class too (Ackermann on https://github.com/ignitionrobotics/ign-gazebo/pull/618/), so it's good to move it to a common place where it can be reused by various controllers instead of copied. Ignition Math feels like a good place to put this.

The class had originally been copied from `ros_control` without much testing. While porting it to `ign-math`, I added tests and noticed that the jerk limit calculation was mistaken and is being fixed on https://github.com/ros-controls/ros_controllers/pull/388. I believe the version on this PR should be correct.

Some other changes:

* Made limits +1inf by default
* Using `chrono` for time

## Checklist
- [x] Added tests


ahcorde (approved):

include/ignition/math/SpeedLimiter.hh
```diff
@@ -0,0 +1,147 @@
...
+    public: SpeedLimiter(
+        bool   _hasVelocityLimits     = false,
+        bool   _hasAccelerationLimits = false,
+        bool   _hasJerkLimits         = false,
+        double _minVelocity           = -INF_D,
+        double _maxVelocity           = INF_D,
+        double _minAcceleration       = -INF_D,
+        double _maxAcceleration       = INF_D,
+        double _minJerk               = -INF_D,
+        double _maxJerk               = INF_D);
```
scpeters, line 86: I'm not a huge fan of APIs that use big long lists of values like this because I usually forget and have to look up the order in which they should be used.

I wonder if we need the `bool` values, since the min and max values now default to unlimited? We could create private `bool` values in the constructor by inspecting the min and max values to see if any are finite.

Aside from that, I don't have any better suggestions for the API

include/ignition/math/SpeedLimiter.hh
```diff
@@ -0,0 +1,147 @@
...
+                         double _v0,
+                         double _v1,
+                         std::chrono::steady_clock::duration _dt) const;
+
+    /// \brief Limit the velocity.
+    /// \param [in, out] _v Velocity to limit [m/s].
+    /// \return Limiting factor, which is (out_v / in_v).
+    public: double LimitVelocity(double &_v) const;
+
+    /// \brief Limit the acceleration.
```
scpeters, line 108: "Limit the acceleration using a first-order backward difference method"

include/ignition/math/SpeedLimiter.hh
```diff
@@ -0,0 +1,147 @@
...
+    /// \param [in, out] _v  Velocity [m/s].
+    /// \param [in] _v0 Previous velocity [m/s].
+    /// \param [in] _dt Time step.
+    /// \return Limiting factor, which is (out_v / in_v).
+    public: double LimitAcceleration(
+        double &_v,
+        double _v0,
+        std::chrono::steady_clock::duration _dt) const;
+
+    /// \brief Limit the jerk.
```
scpeters, line 118: "Limit the jerk using a second-order backward difference method"

include/ignition/math/SpeedLimiter.hh
```diff
@@ -0,0 +1,147 @@
...
+                         std::chrono::steady_clock::duration _dt) const;
+
+    /// \brief Limit the velocity.
+    /// \param [in, out] _v Velocity to limit [m/s].
+    /// \return Limiting factor, which is (out_v / in_v).
+    public: double LimitVelocity(double &_v) const;
+
+    /// \brief Limit the acceleration.
+    /// \param [in, out] _v  Velocity [m/s].
+    /// \param [in] _v0 Previous velocity [m/s].
```
scpeters, line 110: how does `_vPrevious` sound?

include/ignition/math/SpeedLimiter.hh
```diff
@@ -0,0 +1,147 @@
...
+    /// \return Limiting factor, which is (out_v / in_v).
+    public: double LimitAcceleration(
+        double &_v,
+        double _v0,
+        std::chrono::steady_clock::duration _dt) const;
+
+    /// \brief Limit the jerk.
+    /// \param [in, out] _v Velocity to limit [m/s].
+    /// \param [in] _v0 Previous velocity to v  [m/s].
+    /// \param [in] _v1 Previous velocity to v0 [m/s].
```
scpeters, line 121: how about `_v0` -> `_vPrevious1` and `_v1` -> `_vPrevious2`?

src/SpeedLimiter.cc
```diff
@@ -0,0 +1,201 @@
...
+//////////////////////////////////////////////////
+double SpeedLimiter::LimitJerk(double &_v, double _v0, double _v1,
+    std::chrono::steady_clock::duration _dt) const
+{
+  const double tmp = _v;
+
+  if (this->dataPtr->hasJerkLimits)
+  {
+    const double dv  = _v  - _v0;
+    const double dv0 = _v0 - _v1;
```
scpeters, line 184: `dv0` -> `dvPrevious`

src/SpeedLimiter.cc
```diff
@@ -0,0 +1,201 @@
...
+
+    const double dt2 = std::chrono::duration<double>(_dt).count() *
+        std::chrono::duration<double>(_dt).count();
+
+    const double daMin = this->dataPtr->minJerk * dt2;
+    const double daMax = this->dataPtr->maxJerk * dt2;
+
+    const double da = ignition::math::clamp(dv - dv0, daMin, daMax);
```
scpeters, line 192: I think `da` could be renamed `dvFromJerk` (and `dvFromJerkMin` and `dvFromJerkMax` likewise)

include/ignition/math/SpeedLimiter.hh
chapulina: I updated all the variables to match the naming I was using on the tests.

src/SpeedLimiter.cc
chapulina: I think the new implementation looks clearer and we don't need this specific variable anymore.

chapulina:
@scpeters , I refactored the API, implementation and naming conventions. I think the class is easier to use and the math is more straight forward now. Let me know what you think.

include/ignition/math/SpeedLimiter.hh
```diff
@@ -55,77 +33,100 @@ inline namespace IGNITION_MATH_VERSION_NAMESPACE {
...
+
+    /// \brief Set maximum velocity limit in m/s, usually >= 0.
+    /// \param[in] _lim Maximum velocity.
+    public: void SetMaxVelocity(double _lim);
+
+    /// \brief Get maximum velocity limit, defaults to positive infinity.
+    /// \return Maximum velocity.
+    public: double MaxVelocity() const;
+
+    /// \brief Set minimum acceleration limit in m/s, usually <= 0.
```
scpeters, line 61: m/s^2

include/ignition/math/SpeedLimiter.hh
```diff
@@ -55,77 +33,100 @@ inline namespace IGNITION_MATH_VERSION_NAMESPACE {
...
+    /// \param[in] _lim Minimum acceleration.
+    public: void SetMinAcceleration(double _lim);
+
+    /// \brief Get minimum acceleration limit, defaults to negative infinity.
+    /// \return Minimum acceleration.
+    public: double MinAcceleration() const;
+
+    /// \brief Set maximum acceleration limit in m/s, usually >= 0.
```
scpeters, line 69: m/s^2

include/ignition/math/SpeedLimiter.hh
```diff
@@ -55,77 +33,100 @@ inline namespace IGNITION_MATH_VERSION_NAMESPACE {
...
+    /// \param[in] _lim Maximum acceleration.
+    public: void SetMaxAcceleration(double _lim);
+
+    /// \brief Get maximum acceleration limit, defaults to positive infinity.
+    /// \return Maximum acceleration.
+    public: double MaxAcceleration() const;
+
+    /// \brief Set minimum jerk limit in m/s, usually <= 0.
```
scpeters, line 77: m/s^3

include/ignition/math/SpeedLimiter.hh
```diff
@@ -55,77 +33,100 @@ inline namespace IGNITION_MATH_VERSION_NAMESPACE {
...
+    /// \param[in] _lim Minimum jerk.
+    public: void SetMinJerk(double _lim);
+
+    /// \brief Get minimum jerk limit, defaults to negative infinity.
+    /// \return Minimum jerk.
+    public: double MinJerk() const;
+
+    /// \brief Set maximum jerk limit in m/s, usually >= 0.
```
scpeters, line 85: m/s^3

include/ignition/math/SpeedLimiter.hh
```diff
@@ -55,77 +33,100 @@ inline namespace IGNITION_MATH_VERSION_NAMESPACE {
...
+    /// \param[in] _lim Maximum jerk.
+    public: void SetMaxJerk(double _lim);
+
+    /// \brief Get maximum jerk limit, defaults to positive infinity.
+    /// \return Maximum jerk.
+    public: double MaxJerk() const;
+
+    /// \brief Limit velocity, acceleration and jerk.
+    /// \param [in, out] _vel Velocity to limit [m/s].
+    /// \param [in] _prevVel Previous velocity to v  [m/s].
```
scpeters, line 95: Previous velocity to *_vel*

include/ignition/math/SpeedLimiter.hh
```diff
@@ -55,77 +33,100 @@ inline namespace IGNITION_MATH_VERSION_NAMESPACE {
...
+    /// \param [in] _prevPrevVel Previous velocity to prevVel [m/s].
```
scpeters, line 96: add `_` to `prevVel` I think

include/ignition/math/SpeedLimiter.hh
```diff
@@ -55,77 +33,100 @@ inline namespace IGNITION_MATH_VERSION_NAMESPACE {
...
     /// \param [in] _dt Time step.
-    /// \return Limiting factor, which is (out_v / in_v).
-    public: double Limit(double &_v,
-                         double _v0,
-                         double _v1,
+    /// \return Limiting factor, which is (out _vel - in _vel).
```
scpeters, line 98: I wouldn't call it a limiting factor any more since it's not a ratio; I would call it the difference between the output and input values of `_vel` (output - input)

include/ignition/math/SpeedLimiter.hh
```diff
@@ -55,77 +33,100 @@ inline namespace IGNITION_MATH_VERSION_NAMESPACE {
...
+                         double _prevVel,
+                         double _prevPrevVel,
                          std::chrono::steady_clock::duration _dt) const;
 
     /// \brief Limit the velocity.
-    /// \param [in, out] _v Velocity to limit [m/s].
-    /// \return Limiting factor, which is (out_v / in_v).
-    public: double LimitVelocity(double &_v) const;
+    /// \param [in, out] _vel Velocity to limit [m/s].
+    /// \return Limiting factor, which is (out _vel - in _vel).
```
scpeters, line 106: repeat comment about Limiting factor

include/ignition/math/SpeedLimiter.hh
```diff
@@ -55,77 +33,100 @@ inline namespace IGNITION_MATH_VERSION_NAMESPACE {
...
+    public: double LimitVelocity(double &_vel) const;
 
     /// \brief Limit the acceleration.
-    /// \param [in, out] _v  Velocity [m/s].
-    /// \param [in] _v0 Previous velocity [m/s].
+    /// \param [in, out] _vel  Velocity [m/s].
+    /// \param [in] _prevVel Previous velocity [m/s].
     /// \param [in] _dt Time step.
-    /// \return Limiting factor, which is (out_v / in_v).
+    /// \return Limiting factor, which is (out _vel - in _vel).
```
scpeters, line 114: repeat comment about Limiting factor

include/ignition/math/SpeedLimiter.hh
```diff
@@ -55,77 +33,100 @@ inline namespace IGNITION_MATH_VERSION_NAMESPACE {
...
     /// \brief Limit the jerk.
-    /// \param [in, out] _v Velocity to limit [m/s].
-    /// \param [in] _v0 Previous velocity to v  [m/s].
-    /// \param [in] _v1 Previous velocity to v0 [m/s].
+    /// \param [in, out] _vel Velocity to limit [m/s].
+    /// \param [in] _prevVel Previous velocity to v  [m/s].
+    /// \param [in] _prevPrevVel Previous velocity to prevVel [m/s].
     /// \param [in] _dt Time step.
-    /// \return Limiting factor, which is (out_v / in_v).
+    /// \return Limiting factor, which is (out _vel - in _vel).
```
scpeters, line 125: repeat comment about Limiting factor

src/SpeedLimiter.cc
```diff
@@ -97,105 +44,152 @@ class ignition::math::SpeedLimiterPrivate
...
+{
+  return this->dataPtr->maxJerk;
+}
+
+//////////////////////////////////////////////////
+double SpeedLimiter::Limit(double &_vel, double _prevVel, double _prevPrevVel,
     std::chrono::steady_clock::duration _dt) const
 {
-  const double tmp = _v;
+  const double vDesired = _vel;
```
scpeters, line 131: I think `velInput` or `velInitial` or `velUnclamped` make more sense than `vDesired`. this class is just a speed limiter, not a controller with desired values. If we do rename this, we should do so for the other `vDesired`, `accDesired`, and `jerkDesired` variables as well

src/SpeedLimiter.cc
```diff
@@ -97,105 +44,152 @@ class ignition::math::SpeedLimiterPrivate
...
-        std::chrono::duration<double>(_dt).count();
+  const double dtSec = std::chrono::duration<double>(_dt).count();
 
-    const double dv = ignition::math::clamp(_v - _v0, dvMin, dvMax);
+  if (equal(dtSec, 0.0))
+    return 0.0;
 
-    _v = _v0 + dv;
-  }
+  const double accDesired = (_vel - _prevVel) / dtSec;
```
scpeters, line 162: I think the code is clearer to read this way, but I believe it may have been written the other way to avoid a division operation and divide by 0 concerns

src/SpeedLimiter_TEST.cc
```diff
@@ -0,0 +1,222 @@
...
+  double minJerk = -1.0;
+  double maxJerk = 1.0;
+
+  SpeedLimiter limiter;
+  limiter.SetMinVelocity(minVel);
+  limiter.SetMaxVelocity(maxVel);
+  limiter.SetMinAcceleration(minAcc);
+  limiter.SetMaxAcceleration(maxAcc);
+  limiter.SetMinJerk(minJerk);
+  limiter.SetMaxJerk(maxJerk);
```
scpeters, line 191: we could improve the coverage by testing the accessors right after these lines

src/SpeedLimiter.cc
```diff
@@ -0,0 +1,195 @@
...
+//////////////////////////////////////////////////
+double SpeedLimiter::LimitAcceleration(double &_vel, double _prevVel,
+    std::chrono::steady_clock::duration _dt) const
+{
+  const double vDesired = _vel;
+
+  const double dtSec = std::chrono::duration<double>(_dt).count();
+
+  if (equal(dtSec, 0.0))
+    return 0.0;
```
scpeters, line 160: it should be straightforward to add coverage for `_dt` == 0

scpeters (commented):
sorry for the delay in getting back to this. thanks for addressing my earlier comments; I have a few more now

src/SpeedLimiter.cc
chapulina: Yeah I agree that this may have been the motivation. One lesson I took from the fact that the math had been wrong on the `ros_control` implementation all this time is the value of readability. I'm leaning towards leaving the math straightforward so it's easier to debug in the future. I added tests for the `dt == 0` case, so I think we're covered.

scpeters (approved):

PR #206, opened 2021-04-19
👩‍🌾 Prevent -0 with out stream operator

### LINKED ISSUE #161: arm64 test failures during debbuild
There were some `arm64` test failures when building 6.6.0 debs that were successful when building 6.5.0.

* [Build Status](https://build.osrfoundation.org/job/ign-math6-debbuilder/675/) https://build.osrfoundation.org/job/ign-math6-debbuilder/675/console
* [Build Status](https://build.osrfoundation.org/job/ign-math6-debbuilder/682/) https://build.osrfoundation.org/job/ign-math6-debbuilder/682/console
* [Build Status](https://build.osrfoundation.org/job/ign-math6-debbuilder/684/) https://build.osrfoundation.org/job/ign-math6-debbuilder/684/console

~~~
[ RUN      ] PoseTest.OperatorStreamOut
/var/lib/jenkins/workspace/ign-math6-debbuilder/build/ignition-math-6.6.0/src/Pose_TEST.cc:155: Failure
Expected equality of these values:
  stream.str()
    Which is: "0.1 1.2 2.3 -0 0.1 1"
  "0.1 1.2 2.3 0 0.1 1"
[  FAILED  ] PoseTest.OperatorStreamOut (11 ms)
~~~

I'm not sure what has changed to cause this because the difference between 6.5.0 and 6.6.0 is only https://github.com/ignitionrobotics/ign-math/pull/158, which is unrelated to this test failure.

chapulina:
# 🦟 Bug fix

Fixes #161

## Summary

Implemented @azeey 's suggestion to check for zeroes and explicitly output `0` to avoid `-0`.

I also had to add a const version of `Color::operator[]` so it could be used by the operator.

## Checklist
- [x] Added tests (all existing tests should still pass)

https://github.com/osrf/buildfarmer/issues/181

### DIFF
--- a/include/ignition/math/Color.hh
+++ b/include/ignition/math/Color.hh
@@ -74,8 +74,8 @@ namespace ignition
 
       /// \brief Constructor
       /// \param[in] _r Red value (range 0 to 1)
-      /// \param[in] _g Green value (range 0 to 1
-      /// \param[in] _b Blue value (range 0 to 1
+      /// \param[in] _g Green value (range 0 to 1)
+      /// \param[in] _b Blue value (range 0 to 1)
       /// \param[in] _a Alpha value (0=transparent, 1=opaque)
       public: Color(const float _r, const float _g, const float _b,
                   const float _a = 1.0);
@@ -126,10 +126,18 @@ namespace ignition
       public: Color &operator=(const Color &_pt);
 
       /// \brief Array index operator
-      /// \param[in] _index Color component index(0=red, 1=green, 2=blue)
+      /// \param[in] _index Color component index(0=red, 1=green, 2=blue,
+      /// 3=alpha)
       /// \return r, g, b, or a when _index is 0, 1, 2 or 3. A NAN_F value is
       /// returned if the _index is invalid
       public: float operator[](const unsigned int _index);
+
+      /// \brief Array index operator, const version
+      /// \param[in] _index Color component index(0=red, 1=green, 2=blue,
+      /// 3=alpha)
+      /// \return r, g, b, or a when _index is 0, 1, 2 or 3. A NAN_F value is
+      /// returned if the _index is invalid
+      public: float operator[](const unsigned int _index) const;
 
       /// \brief Get as uint32 RGBA packed value
       /// \return the color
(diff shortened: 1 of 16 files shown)


include/ignition/math/Color.hh
```diff
@@ -131,6 +131,12 @@ namespace ignition
       /// returned if the _index is invalid
       public: float operator[](const unsigned int _index);
 
+      /// \brief Array index operator, const version
+      /// \param[in] _index Color component index(0=red, 1=green, 2=blue)
```
scpeters, line 135: nit: add ", 3=alpha"?

scpeters (commented):
I added a test case in e3c7b81a00d2f58a44f02556a3cbfaf97a423a1d that fails on `ign-math6`: printing a `Pose3::Zero` object results in `0 0 0 0 -0 0`. With this branch we can expect all `0` values

there are some behavior changes by setting precision to 6 for all our output streams, though it does make them more consistent

chapulina:
> I added a test case in e3c7b81 that fails on ign-math6

Thank you for the test case. I'm on he fence here about whether it's a good idea to get this PR in. Consistency is good, but behavior change, even if it's almost harmless, can be disruptive for users who have been handling the current behaviour. I can target this at `main` if you prefer.

include/ignition/math/Color.hh
```diff
@@ -126,11 +126,17 @@ namespace ignition
...
 
       /// \brief Array index operator
-      /// \param[in] _index Color component index(0=red, 1=green, 2=blue)
+      /// \param[in] _index Color component index(0=red, 1=green, 2=blue, 3=alpha)
       /// \return r, g, b, or a when _index is 0, 1, 2 or 3. A NAN_F value is
       /// returned if the _index is invalid
       public: float operator[](const unsigned int _index);
 
+      /// \brief Array index operator, const version
+      /// \param[in] _index Color component index(0=red, 1=green, 2=blue, 3=alpha)
```
scpeters, line 135: codecheck in github actions is complaining about line lengths on two lines in this file

chapulina:
> There are tests in libsdformat that expect -0 to be output. I am okay with updating them to 0 immediately after this is merged, but I just wanted to point out the behavior change.

Ouch that's good to know. I think we should retarget this to `main`. Ignition Math 6 is used across various projects and has been around for over 2 years. I think it's safe to assume that this will break users while bringing little benefit.

chapulina:
humm I don't know what's happening on CI:

```
stderr: error: unable to create file tools/check_test_ran.py (Permission denied)
error: unable to create file tools/code_check.sh (Permission denied)
fatal: cannot create directory at 'tools/cppcheck_rules': Permission denied
```

Seems related to #211. Any ideas, @j-rivero ?

j-rivero:
> Seems related to #211. Any ideas, @j-rivero ?

For some reason the tools/ directory was under root permissions (could be a build finished unsuccessful). I've change it to jenkins and things are working as expected.

j-rivero (commented):
Minor suggestion for the Migration document. Looks good to me. I was trying to measure the impact on performance of changes by playing with `time` command but my results are pretty inconsistent. Given that it will only affect to the output I'm not that worried.

j-rivero (approved):

chapulina:
> I was trying to measure the impact on performance of changes by playing with time command but my results are pretty inconsistent. Given that it will only affect to the output I'm not that worried.

That's a good idea. This may actually impact Gazebo, because the stream operators are used for serialization, and that happens often. So let's keep an eye on that :eye:

PR #208, opened 2021-04-23
Clean up a few SWIG-related warnings

mjcarroll:
* Set policies when relevant
* Suppress warnings on SWIG-generated files


chapulina (approved):
Looks reasonable to me, no reason to track those warnings if we can't fix them.

Here's some documentation on [CMP0078](https://cmake.org/cmake/help/git-stage/policy/CMP0078.html) as reference.

scpeters (approved):

PR #219, opened 2021-08-12
Arjo/volume below

arjo129:

# 🎉 New feature

Adds a  `VolumeBelow(Plane<T> &_plane)` function to some of the shapes. This is useful for simulating surface behavior when the vehicle is buoyant.

## Summary

Currently the following shapes have a `VolumeBelow`:
* Box
* Sphere

The sphere calculation is relatively simple formula can be found [here](https://en.wikipedia.org/wiki/Spherical_cap).
The box calculation is relatively difficult. However the essence of it is we find out the points at which the plane intersects with the plane itself. We then reconstruct a triangle mesh using the resulting vertices. Subsequently we use the generic [triangle mesh volume formula](http://chenlab.ece.cornell.edu/Publication/Cha/icip01_Cha.pdf).

*TODO*:
* Cylinders
Cylinder slices are relatively simple to calculate using Cylinder wedges. 

## Checklist
- [ ] Added tests


include/ignition/math/Box.hh
```diff
@@ -166,6 +180,10 @@ namespace ignition
 
       /// \brief The box's material.
       private: ignition::math::Material material;
+
+      /// \brief Get intersection between a plane and the box.
+      private: std::vector<Vector3<Precision>> GetIntersections(
```
chapulina, line 185: Style nit:

```suggestion
      private: std::vector<Vector3<Precision>> Intersections(
```

include/ignition/math/Box.hh
```diff
@@ -128,6 +131,17 @@ namespace ignition
...
       public: Precision Volume() const;
 
+      /// \brief Get the volume of the box below a plane.
+      /// \param[in] _plane The plane which cuts the box.
+      /// \return Volume below the plane in m^3.
+      public: Precision VolumeBelow(const Plane<Precision> &_plane) const;
+
+      /// \brief Center of volume below the plane. This is useful when
+      /// calculating where the buoyancy should be applied.
+      /// \param[in] _plane - the plane which slices the sphere.
```
chapulina, line 141:
```suggestion
      /// \param[in] _plane The plane which slices the sphere.
```

include/ignition/math/Cylinder.hh
```diff
@@ -130,6 +133,10 @@ namespace ignition
       /// \return Volume of the cylinder in m^3.
       public: Precision Volume() const;
 
+      /// \brief Get the volume of the cylinder in m^3.
+      /// \return Volume of the cylinder in m^3.
```
chapulina, line 137: Document `_plane`

include/ignition/math/Line3.hh
```diff
@@ -229,6 +229,15 @@ namespace ignition
         return true;
       }
 
+      /// \brief Calculate distance between line and slope
+      /// \param[in] _pt point which we are measuring distance to.
+      /// \returns Distance from point to line.
+      public: T Distance(const Vector3<T> _pt)
```
chapulina, line 235:
```suggestion
      public: T Distance(const Vector3<T> &_pt)
```

include/ignition/math/Line3.hh
```diff
@@ -229,6 +229,15 @@ namespace ignition
         return true;
       }
 
+      /// \brief Calculate distance between line and slope
```
chapulina, line 232: Did you mean

```suggestion
      /// \brief Calculate distance between line and point
```

include/ignition/math/Plane.hh
```diff
@@ -119,6 +122,76 @@ namespace ignition
         return this->normal.Dot(_point) - this->d;
       }
 
+      /// \brief Given x,y point find corresponding Z point on plane.
+      /// \param[in] x - 2d x coordinate.
+      /// \param[in] y - 2d y coordinate.
+      /// \param[in] eps - The value to consider zero. Defaults to 1e-16
+      /// \return coincident point.
+      public: Vector3<T> GetPointOnPlane(const T x,
```
chapulina, line 130: * Style

    ```suggestion
      public: Vector3<T> PointOnPlane(const T x,
    ```

* Just reading the documentation, it took me a while to understand what the function was doing. Maybe saying something along the lines of "given a point on the horizontal XY plane (i.e. Z == 0), find that point projected on the plane"
* Can we use `Vector2` instead of separate x/y?
* Reminder to update style:
    * prefix inputs with `_`
    * use `camelCase`

include/ignition/math/Plane.hh
```diff
@@ -119,6 +122,76 @@ namespace ignition
...
+
+      /// \brief Get the intersection of a line with the plane
+      /// given the line's gradient and a point in parametrized space.
+      /// \param[in] _point A point that lies on a line.
+      /// \param[in] _gradient The gradient of the line.
+      /// \return The point of intersection. std::nullopt if the line is
+      /// parrallel to the plane.
+      public: std::optional<Vector3<T>> Intersect(
+        const Vector3<T> &_point,
+        const Vector3<T> &_gradient) const
```
chapulina, line 149: Can we use `Line3` instead of a point + gradient? That should be more well defined.

include/ignition/math/Sphere.hh
```diff
@@ -91,6 +92,18 @@ namespace ignition
...
 
+      /// \brief Get the volume of sphere below a given plane in m^3.
+      /// It is assumed that the center of the sphere is on the origin
+      /// \param[in] _plane - the plane which slices this sphere.
+      /// \return Volume below the sphere in m^3.
+      public: Precision VolumeBelow(const Planed &_plane) const;
+
+      /// \brief Center of volume below the plane. This is useful when
+      /// calculating where the buoyancy should be applied.
+      /// \param[in] _plane - the plane which slices the sphere.
```
chapulina, line 103:
```suggestion
      /// \param[in] _plane The plane which slices the sphere.
```

chapulina (commented):
I did a high-level first pass. We should make sure all added functions have unit tests.

> TODO:   Cylinders

@arjo129 , I noticed you opened this PR for review, but this TODO is still listed and I noticed that you didn't add a `Cylinder::CenterOfVolumeBelow` function. Are you planning to add one?

arjo129:
I can do that later in a separate PR after the fortress release.

include/ignition/math/Box.hh
arjo129: Resolved in 9ed6254

include/ignition/math/Box.hh
arjo129: Resolved in 9ed6254

include/ignition/math/Cylinder.hh
arjo129: Removed in 9ed6254

arjo129:
This is for tracking my own progress:

Function Name | Unit Tested
-- | --
Box::VolumeBelow | Yes
Box::CenterOfVolumeBelow | No
Box::Intersections | No
Line::Distance | No
Plane::PointOnPlane | Yes
Plane::Intersect | Yes
Plane::Volume | No
Sphere::VolumeBelow | Yes
Sphere::CenterOfVolumeBelow | Yes
Vector3::Project | No

include/ignition/math/Line3.hh
arjo129: Resolved in 7f1f1c8

include/ignition/math/Line3.hh
arjo129: Resolved in 7f1f1c8

include/ignition/math/Plane.hh
arjo129: `Line3`s have a start and an end the `Intersect()` routine would look different from the point of view of a `Line3` vs an infinite line.

include/ignition/math/Sphere.hh
arjo129: Resolved in 633dfee

include/ignition/math/Plane.hh
arjo129: Removed in 45920e1 as this is not used anywhere and does not bring any happiness. Its probably from an older algorithm I was trying.

chapulina (changes requested):
I'm still reviewing it, see some suggestions on #240

include/ignition/math/Vector3.hh
```diff
@@ -264,6 +264,14 @@ namespace ignition
         return n.Normalize();
       }
 
+      /// \brief Get Projection of another onto this vector
+      /// \param[in] _v the vector
+      /// \return the projection
+      public: T Project(const Vector3<T> &_v) const
+      {
+        return this->Dot(_v) / this->Length();
```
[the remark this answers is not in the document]
arjo129: Nope. Not necessarily https://en.wikipedia.org/wiki/Vector_projection .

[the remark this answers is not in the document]
chapulina: What part says that the projection could be longer? The scalar projection is the length of the vector multiplied by the cosine of the angle between the 2 vectors, which is always <= 1, right?

include/ignition/math/detail/Sphere.hh
```diff
@@ -95,6 +95,58 @@ T Sphere<T>::Volume() const
...
+    // sphere is completely below plane
+    return Volume();
+  }
+  else if(dist > r)
+  {
+    // sphere is completely above plane
+    return 0;
+  }
+
+  auto h = r + dist;
```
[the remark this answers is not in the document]
chapulina: See #240 for the change with tests

include/ignition/math/Plane.hh
```diff
@@ -119,6 +122,26 @@ namespace ignition
         return this->normal.Dot(_point) - this->d;
       }
 
+      /// \brief Get the intersection of a line with the plane
+      /// given the line's gradient and a point in parametrized space.
+      /// \param[in] _point A point that lies on a line.
+      /// \param[in] _gradient The gradient of the line.
+      /// \return The point of intersection. std::nullopt if the line is
+      /// parrallel to the plane.
+      public: std::optional<Vector3<T>> Intersect(
```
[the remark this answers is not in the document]
arjo129: I agree, but I need a Plane Object without ends for the Buoyancy plugin.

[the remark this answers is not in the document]
chapulina: The class is not well documented, but I'd say we can treat a plane with size 0,0 as infinite (i.e. the default size)

include/ignition/math/Vector3.hh
```diff
@@ -264,6 +264,14 @@ namespace ignition
         return n.Normalize();
       }
 
+      /// \brief Get Projection of another onto this vector
+      /// \param[in] _v the vector
+      /// \return the projection
+      public: T Project(const Vector3<T> &_v) const
```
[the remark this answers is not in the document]
arjo129: This is not a ScalarProjection though,,, Its the length of the Vector projection. Combing through the code base its not used anywhere and can be removed.

include/ignition/math/Vector3.hh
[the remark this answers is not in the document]
arjo129: I should have been clearer in naming. Its not a Scalar projection.its the length of the Vector projection. A search of the codebase also shows I no longer use this function. Perhaps I should remove it.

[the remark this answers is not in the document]
chapulina: I'm ok removing it if it's not needed now, it can always be added later :+1:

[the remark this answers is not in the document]
arjo129: Apologies, this class is used in calculating the box volume. And it does need to be the length of the VectorProjection.

include/ignition/math/detail/Sphere.hh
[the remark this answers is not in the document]
arjo129: Thanks for spotting this

include/ignition/math/Vector3.hh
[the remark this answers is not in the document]
chapulina: I think I got confused with the test expectations because I was thinking that on that 2nd expectation, `v2` was projected onto `v3`, so it was weird that the projection had length 3 while the original vector had length 1.

https://github.com/ignitionrobotics/ign-math/blob/cb912344a4974ccfc7e4ae2938518dcc3cd63c06/src/Vector3_TEST.cc#L441-L447

The order of operations also makes sense for me now, sorry for the noise.

> Its not a Scalar projection.its the length of the Vector projection

I _think_ those should be the same, according to Wikipedia:

"_The vector projection of a on b is a vector whose magnitude is the scalar projection of a on b with the same direction as b."_

include/ignition/math/Plane.hh
```diff
@@ -119,6 +122,26 @@ namespace ignition
...
+      /// given the line's gradient and a point in parametrized space.
+      /// \param[in] _point A point that lies on the line.
+      /// \param[in] _gradient The gradient of the line.
+      /// \return The point of intersection. std::nullopt if the line is
+      /// parrallel to the plane.
+      public: std::optional<Vector3<T>> Intersection(
+        const Vector3<T> &_point,
+        const Vector3<T> &_gradient) const
+      {
+        if (abs(this->Normal().Dot(_gradient)) < 1e-6)
```
scpeters, line 135: I would recommend adding a `_tolerance` parameter that defaults to `1e-6` (see some of the APIs in [MassMatrix3.hh](https://github.com/ignitionrobotics/ign-math/blob/b54dce578d15ec24f2008b95d7b9d08f13ba5613/include/ignition/math/MassMatrix3.hh#L475) for similar examples)

include/ignition/math/Vector3.hh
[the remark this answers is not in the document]
scpeters: I'm concerned that this will result in a divide by zero if `this` is a Zero vector

what should this method return in that case?

scpeters (commented):
to simplify the review process, we could split into two PR's, one for sphere and one for box functions. It's not mandatory but could get the sphere portion merged sooner, since it's a bit simpler

include/ignition/math/Vector3.hh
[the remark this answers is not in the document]
arjo129: Hmm technically I don't think Vector Projections are defined for zero length vectors. I could return zero but I'm not sure thats the best way to go around it.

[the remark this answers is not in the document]
scpeters: considering where you are using this method; do you currently check the length of the vector before calling this method?

arjo129:
To be honest, for the project this is targeting, the box method is higher priority than the sphere method.

include/ignition/math/Plane.hh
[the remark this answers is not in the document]
arjo129: Resolved in 9a66163

chapulina (changes requested):
Looking good to me! I have my final suggestions in https://github.com/ignitionrobotics/ign-math/pull/243 for your consideration.

include/ignition/math/Vector3.hh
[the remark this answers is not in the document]
arjo129: Its now used in more than one function. I've added a doc-string warning that ts can return nans.

arjo129:
@chapulina Windows is fixed!! I recommend one more glance at 20d66b1. We have a weird issue with `std::set` and the fact that `Vector3<T>::operator<` is not actually strictly ordered. I've added a `WellOrderedVector` function object.

W.r.t `VectorProjectionLength` while we could wrap it with an `optional`, I don't actually want to copy it everywhere, cause it is used quite a bit. I do feel that a warning in the documentation should suffice. Afterall, we do not check if there are `NaN`s when we do basic operations.

chapulina:
> We have a weird issue with std::set and the fact that Vector3<T>::operator< is not actually strictly ordered.

Thank you for the workaround, LGTM. The only concern I have is that a `detail` struct would be part of the public API, but users shouldn't be interacting with it, so I think it's ok.

Ticketed https://github.com/ignitionrobotics/ign-math/issues/244 to keep track of this.

chapulina (approved):
Thanks for all the iteration! Let's :ship: it! :rocket:

include/ignition/math/detail/Box.hh
```diff
@@ -113,6 +121,155 @@ T Box<T>::Volume() const
...
+
+  return triangles;
+}
+
+/////////////////////////////////////////////////
+template<typename T>
+T Box<T>::VolumeBelow(const Plane<T> &_plane) const
+{
+  auto verticesBelow = this->VerticesBelow(_plane);
+  if (verticesBelow.size() == 0)
```
scpeters, line 181: nit: use `if (verticesBelow.empty())`

include/ignition/math/Vector3.hh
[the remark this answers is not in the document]
scpeters: >> Another option is not adding this function to `Vector3` for now and just calculating it directly within that function for now.

> Its now used in more than one function. I've added a doc-string warning that ts can return nans.

I see that it is used in multiple places, but I don't see any logic that checks for NaNs. Since this function is currently only one line, I would prefer to redact this API and duplicate this one line of code with context-specific error-checking in the two functions where it is currently used

include/ignition/math/Line3.hh
```diff
@@ -229,6 +229,32 @@ namespace ignition
...
+
+        // Point is projected beyond pt1
+        if (ptTo1.Dot(line) >= 0.0)
+        {
+          return ptTo1.Length();
+        }
+
+        // Distance to point projected onto line
+        auto d = (ptTo0).Cross(line);
+        return d.Length() / (line).Length();
```
scpeters, line 255: the other APIs in this class accept an `_epsilon` parameter with a default value of `1e-6`. I think we should do something similar to avoid the possibility of dividing by `0` here

include/ignition/math/Vector3.hh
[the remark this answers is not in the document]
arjo129: Removed the function and inlined it.

include/ignition/math/Line3.hh
arjo129: Resolved in 72d7779.

arjo129:
@scpeters I've addressed your feedback and removed the controversial function.

include/ignition/math/Plane.hh
```diff
@@ -157,8 +157,8 @@ namespace ignition
         Vector3<T> rotatedXAxis = rotation * Vector3<T>::UnitX;
         Vector3<T> rotatedYAxis = rotation * Vector3<T>::UnitY;
 
-        auto xBasis = rotatedXAxis.VectorProjectionLength(intersection);
-        auto yBasis = rotatedYAxis.VectorProjectionLength(intersection);
+        auto xBasis = rotatedXAxis.Dot(intersection);
+        auto yBasis = rotatedYAxis.Dot(intersection);
```
chapulina, line 161: This was dividing by length before

arjo129: Yes. But the length of the unit vectors is 1.

include/ignition/math/Line3.hh
```diff
@@ -251,8 +252,13 @@ namespace ignition
         }
 
         // Distance to point projected onto line
-        auto d = (ptTo0).Cross(line);
-        return d.Length() / (line).Length();
+        auto d = ptTo0.Cross(line);
+
+        // If line has length zero assume its a point.
+        if(line.Length() <= _epsilon)
```
chapulina, line 258: Is there a way to get here? I tried adding these test cases but they're already covered by the dot product takes care of that already.

```diff
diff --git a/src/Line3_TEST.cc b/src/Line3_TEST.cc
index de2c531..c95a2a3 100644
--- a/src/Line3_TEST.cc
+++ b/src/Line3_TEST.cc
@@ -290,6 +290,26 @@ TEST(Line3Test, DistanceToPoint)
     math::Vector3d point(-5, -3, -8);
     EXPECT_DOUBLE_EQ(line.Distance(point), point.Distance(pointB));
   }
+
+  // Zero-sized line
+  pointA.Set(0, 0, 0);
+  line.Set(pointA, pointA);
+
+  // Point on the line
+  {
+    math::Vector3d point(0, 0, 0);
+    EXPECT_DOUBLE_EQ(line.Distance(point), 0.0);
+  }
+
+  // Points projected beyond the line's ends
+  {
+    math::Vector3d point(2, 2, 3);
+    EXPECT_DOUBLE_EQ(line.Distance(point), point.Distance(pointA));
+  }
+  {
+    math::Vector3d point(-5, -3, -8);
+    EXPECT_DOUBLE_EQ(line.Distance(point), point.Distance(pointA));
+  }
 }
 
 /////////////////////////////////////////////////
```

@scpeters , do you think the tolerance is really necessary?

arjo129: I agree with you upon closer inspection. This looks like dead code. Should I revert it?

include/ignition/math/Plane.hh
chapulina: Ah yes, I hadn't noticed that

include/ignition/math/Line3.hh
scpeters: you're right; it doesn't ever reach that point

include/ignition/math/Line3.hh
```diff
@@ -229,6 +229,32 @@ namespace ignition
...
+
+        // Point is projected beyond pt1
+        if (ptTo1.Dot(line) >= 0.0)
+        {
+          return ptTo1.Length();
+        }
+
+        // Distance to point projected onto line
+        auto d = ptTo0.Cross(line);
+        return d.Length() / line.Length();
```
scpeters, line 255: we can add a comment that `line.Length()` mush be `> 0` at this point since the code would have already `return`ed if it was `== 0`. We could even add an assert if we wanted to be formal about it

~~~
auto lineLength = line.Length();
assert(lineLength > 0);
return d.Length() / lineLength;
~~~

src/Line3_TEST.cc
```diff
@@ -230,6 +230,68 @@ TEST(Line3Test, Distance)
   EXPECT_FALSE(line.Distance(math::Line3d(2, 0, 0, 2, 1, 0), result));
 }
 
+/////////////////////////////////////////////////
+TEST(Line3Test, DistanceToPoint)
```
scpeters, line 234: add a test for a degenerate line with identical endpoints

include/ignition/math/Line3.hh
arjo129: Resolved in f547e71

include/ignition/math/detail/Box.hh
```diff
@@ -139,6 +300,55 @@ bool Box<T>::MassMatrix(MassMatrix3<T> &_massMat) const
...
+  // Axes
+  std::vector<Vector3<T>> axes
+  {
+    Vector3<T>{1, 0, 0},
+    Vector3<T>{0, 1, 0},
+    Vector3<T>{0, 0, 1}
+  };
+
+  // There are 12 edges. However, we just need 4 points to describe the
+  // twelve edges.
```
scpeters, line 330: suggested rewording:

"There are 12 edges, which are checked along 3 axes from 4 box corner points."

include/ignition/math/detail/Box.hh
```diff
@@ -113,6 +121,159 @@ T Box<T>::Volume() const
...
+/////////////////////////////////////////////////
+template<typename T>
+T Box<T>::VolumeBelow(const Plane<T> &_plane) const
+{
+  auto verticesBelow = this->VerticesBelow(_plane);
+  if (verticesBelow.empty())
+    return 0;
+
+  auto intersections = this->Intersections(_plane);
+  verticesBelow.merge(intersections);
```
scpeters, line 189: I believe this implicitly uses the `==` operator for `Vector3`, which has a default tolerance of `1e-6`. If tolerances become an issue, we could overload this function with an `_epsilon` parameter in the future, but I don't think that is necessary now

include/ignition/math/detail/Box.hh
```diff
@@ -113,6 +121,159 @@ T Box<T>::Volume() const
...
+
+  // Fit six planes to the vertices in the shape.
```
scpeters, line 191: I don't understand what is happening here

include/ignition/math/detail/Box.hh
```diff
@@ -113,6 +121,159 @@ T Box<T>::Volume() const
...
+  std::vector<std::pair<Triangle3<T>, T>> triangles;
+
+  std::vector<Plane<T>> planes
+  {
+    Plane<T>{Vector3<T>{0, 0, 1}, this->Size().Z()/2},
+    Plane<T>{Vector3<T>{0, 0, -1}, this->Size().Z()/2},
+    Plane<T>{Vector3<T>{1, 0, 0}, this->Size().X()/2},
+    Plane<T>{Vector3<T>{-1, 0, 0}, this->Size().X()/2},
+    Plane<T>{Vector3<T>{0, 1, 0}, this->Size().Y()/2},
+    Plane<T>{Vector3<T>{0, -1, 0}, this->Size().Y()/2},
```
scpeters, line 201: should half of the `this->Size().*` arguments have negative signs? I'm not sure what this is computing

include/ignition/math/detail/Box.hh
```diff
@@ -113,6 +121,159 @@ T Box<T>::Volume() const
   return this->size.X() * this->size.Y() * this->size.Z();
 }
 
+//////////////////////////////////////////////////
+template <typename T>
+std::vector<std::pair<Triangle3<T>, T>> TrianglesInPlane(
```
scpeters, line 126: this helper function could use a doc-string with a high-level explanation of what it does

include/ignition/math/detail/Box.hh
arjo129: Apologies, I think "Reconstruct the cut-box as a triangle mesh" is a better comment.

scpeters: yes, that is more informative

include/ignition/math/detail/Box.hh
arjo129: No since the normals have been negated in half of them.

chapulina: I believe these are the 6 faces of the box. The normals are pointing in opposite directions, so I think it makes sense that the half sizes are all positive.

scpeters: right, I was mistaken about how the plane constant works

include/ignition/math/detail/Box.hh
```diff
@@ -122,10 +122,16 @@ T Box<T>::Volume() const
...
+/// compute the list of triangles which form this polygon. 
+/// \param[in] _plane The plane in which the vertices exist.
+/// \param[in] _vertices The vertices of the polygon.
+/// \return a list of triangles, or and empty vector if _vertices in the _plane
+/// are less than 3.
 template <typename T>
 std::vector<std::pair<Triangle3<T>, T>> TrianglesInPlane(
     const Plane<T> &_plane, IntersectionPoints<T> &_vertices)
-{
+{
```
scpeters: whitespace

include/ignition/math/detail/Box.hh
```diff
@@ -113,6 +121,165 @@ T Box<T>::Volume() const
   return this->size.X() * this->size.Y() * this->size.Z();
 }
 
+//////////////////////////////////////////////////
+/// \brief Given a *convex* polygon described by the verices in a given plane,
+/// compute the list of triangles which form this polygon.
+/// \param[in] _plane The plane in which the vertices exist.
+/// \param[in] _vertices The vertices of the polygon.
+/// \return a list of triangles, or and empty vector if _vertices in the _plane
```
scpeters, line 129: spelling: "or *an* empty vector"

also, can you document the scalar 2nd value in the `std::pair`?

arjo129: b4f9ddc

include/ignition/math/detail/Box.hh
arjo129: Added a todo in f547e71.

include/ignition/math/detail/Box.hh
```diff
@@ -113,6 +121,167 @@ T Box<T>::Volume() const
   return this->size.X() * this->size.Y() * this->size.Z();
 }
 
+//////////////////////////////////////////////////
+/// \brief Given a *convex* polygon described by the verices in a given plane,
+/// compute the list of triangles which form this polygon.
+/// \param[in] _plane The plane in which the vertices exist.
+/// \param[in] _vertices The vertices of the polygon.
+/// \return A vector of triangles and their centroids, or an empty vector
```
scpeters: it looks like the 2nd value in the `std::pair`, which has type `T` is either `+1` or `-1`,  depending on the orientation of the plane. I think that should be documented in the `\return` doc-string

scpeters: you were faster! b4f9ddce0441145a3401581c5f707c43bf71e09e

scpeters (approved):
thanks for your patience with my review! the API looks good to me. I haven't gone over the test cases, but I think it is good for merging once CI is happy

PR #222, opened 2021-08-18
Add Vector(2|3|4)<T>::NaN to easily create invalid vectors 

chapulina:
# 🎉 New feature

## Summary

I often need to create invalid vectors to indicate that something isn't initialized or became invalid. There are ways of expressing this with null pointers or `std::nullopt`, but using NaNs makes it convenient to do without involving the creation and destruction of new vectors.

This also plays well with the existing `IsFinite` and `Correct` functions. Note that [none of this applies to integer types](https://stackoverflow.com/questions/3949457/can-an-integer-be-nan-in-c).

## Test it

Take a look at the added tests.

## Checklist
- [x] Added tests


nkoenig:
@chapulina , do you want to try adding the `nan` types to the corresponding swig `.i` files?

chapulina:
> do you want to try adding the nan types to the corresponding swig .i files?

Done in 1233218

src/ruby/Vector2_TEST.rb
```diff
@@ -257,6 +257,28 @@ def test_length
...
+  def test_nan
+    nanVec = Ignition::Math::Vector2d.NaN
+    assert(!nanVec.IsFinite(),
+           "NaN vector shouldn't be finite")
+    assert(nanVec.X().nan?, "X should be NaN")
+    assert(nanVec.Y().nan?, "Y should be NaN")
+
+    nanVec.Correct()
+    assert(Ignition::Math::Vector2d.Zero == nanVec,
+           "Corrected vector equals zero")
```
azeey, line 270: nit: Should it say "Corrected vector should equal zero"?

azeey (approved):
Looks good. Although I'd argue `std::optional` is more convenient than this for representing uninitialized values. For one, when checking if a variable is nan, you can't compare it against `Vector::NaN` directly. You'd have to use `IsFinite`, which I don't think would be immediately apparent to users. Also, `IsFinite` returns false if any of the members are `nan`, or `inf`, so it could potentially lead to bugs if `Vector::NaN` is used to represent uninitialized vectors.

chapulina:
Thanks for the review, @azeey, I addressed your comments in 564de1c.

I also appreciate the feedback on the motivation for this PR. The specific use case that motivated this was a vector whose value keeps changing, and at times I actually needed to fill it with NaNs. But at other times I've also needed to "flag" that a vector's value was invalid temporarily, and I think it could be useful there as well, but I agree that there are multiple ways of going about this.

PR #224, opened 2021-08-18
Add eigen utils to convert mesh 3d vertices to oriented box

AmrElsersy:
Signed-off-by: AmrElsersy <amrelsersay@gmail.com>

# 🎉 New feature

## Summary
Adding helper functions to convert 3d vertices of a mesh to an oriented bounding box to be used in the bounding box camera 
[rendering#334](https://github.com/ignitionrobotics/ign-rendering/pull/334)

## Checklist
- [ ] Added tests


eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,143 @@
...
+
+        covariance(1, 0) = covariance(0, 1);
+        covariance(2, 0) = covariance(0, 2);
+        covariance(2, 1) = covariance(1, 2);
+
+        return covariance;
+      }
+
+      /// \brief Get the oriented 3d bounding box of a mesh points using PCA
+      inline ignition::math::OrientedBoxd MeshToOrientedBox(
```
adlarkin, line 69: Can you also add reference for the algorithm you used? I believe this is the link you mentioned in the past, right? http://codextechnicanum.blogspot.com/2015/04/find-minimum-oriented-bounding-box-of.html

eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,143 @@
...
+      {
+        math::OrientedBoxd box;
+
+        // Centroid (mean) of the mesh
+        math::Vector3d mean;
+        for (auto point : _mesh)
+          mean += point;
+        mean /= _mesh.size();
+
+        Eigen::Vector3d centroid = convert(mean);
```
adlarkin, line 80:
```suggestion
        Eigen::Vector3d centroid = math::eigen3::convert(mean);
```
Same as my other comment about using the namespace to avoid ambiguity

eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,143 @@
...
+
+        // Covariance
+        Eigen::Matrix3d covariance = computeCovarianceMatrix(_mesh, centroid);
```
adlarkin, line 83:
```suggestion
        Eigen::Matrix3d covariance = computeCovarianceMatrix(_mesh, centroid);
```
The comment here is redundant since the method and variable name explain what is going on here

eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,143 @@
...
+        // The numbers come out the same without it, but The signs are
+        // different and the box doesn't get correctly oriented in some cases.
+        eigenVectorsPCA.col(2) =
+          eigenVectorsPCA.col(0).cross(eigenVectorsPCA.col(1));
+
+        // Transform the original cloud to the origin where the principal
+        // components correspond to the axes.
+        Eigen::Matrix4d projectionTransform(Eigen::Matrix4d::Identity());
+        projectionTransform.block<3, 3> (0, 0) = eigenVectorsPCA.transpose();
+        projectionTransform.block<3, 1> (0, 3) =
```
adlarkin, line 100:
```suggestion
        projectionTransform.block<3, 3>(0, 0) = eigenVectorsPCA.transpose();
        projectionTransform.block<3, 1>(0, 3) =
```

eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,143 @@
...
+          -1.f * (projectionTransform.block<3, 3>(0, 0) * centroid);
```
adlarkin, line 101:
```suggestion
          -1.0f * (projectionTransform.block<3, 3>(0, 0) * centroid);
```

eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,143 @@
...
+
+        Eigen::Vector3d minPoint(INF_I32, INF_I32, INF_I32);
+        Eigen::Vector3d maxPoint(-INF_I32, -INF_I32, -INF_I32);
+
+        // Get the minimum and maximum points of the transformed cloud.
+        for (auto point : _mesh)
```
adlarkin, line 107:
```suggestion
        for (const auto &point : _mesh)
```
Pass by reference to avoid copying a `math::Vector3d` object

eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,143 @@
...
+        {
+          Eigen::Vector4d pt(0, 0, 0, 1);
+          pt.head<3>() = convert(point);
+          Eigen::Vector4d tfPoint = projectionTransform * pt;
+          minPoint = minPoint.cwiseMin(tfPoint.head<3>());
+          maxPoint = maxPoint.cwiseMax(tfPoint.head<3>());
+        }
+
+        const Eigen::Vector3d meanDiagonal = 0.5f*(maxPoint + minPoint);
```
adlarkin, line 116:
```suggestion
        const Eigen::Vector3d meanDiagonal = 0.5f * (maxPoint + minPoint);
```

eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,143 @@
...
+
+        // quaternion is calculated using the eigenvectors (which determines
+        // how the final box gets rotated), and the transform to put the box
+        // in correct location is calculated
+        const Eigen::Quaterniond bboxQuaternion(eigenVectorsPCA);
+        const Eigen::Vector3d bboxTransform =
+          eigenVectorsPCA * meanDiagonal + centroid;
+
+        // xyz xzy ... yxz yzx ... zyx zxy
```
adlarkin, line 125: What's the significance of this comment? It's confusing me :thinking: :grimacing:

AmrElsersy: I was debugging something, will remove it :D

eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,143 @@
...
+      inline ignition::math::OrientedBoxd MeshToOrientedBox(
+        const std::vector<math::Vector3d> &_mesh)
+      {
+        math::OrientedBoxd box;
+
+        // Centroid (mean) of the mesh
+        math::Vector3d mean;
+        for (auto point : _mesh)
+          mean += point;
+        mean /= _mesh.size();
```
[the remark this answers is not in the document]
AmrElsersy: it gives me error 
```
ignition/ign-math/include/ignition/math/Helpers.hh:505:15: error: conversion from ‘int’ to non-scalar type ‘ignition::math::v6::Vector3<double>’ requested
  505 |       T sum = 0;
```

it seems that it is not suitable for double stuff

eigen3/include/ignition/math/eigen3/Util.hh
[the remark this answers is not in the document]
scpeters: +1

eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,143 @@
...
+        }
+
+        covariance(1, 0) = covariance(0, 1);
+        covariance(2, 0) = covariance(0, 2);
+        covariance(2, 1) = covariance(1, 2);
+
+        return covariance;
+      }
+
+      /// \brief Get the oriented 3d bounding box of a mesh points using PCA
```
scpeters, line 68: need Doxygen for parameters and return value

eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,143 @@
...
+#include <ignition/math/eigen3/Conversions.hh>
+#include <ignition/math/OrientedBox.hh>
+
+namespace ignition
+{
+  namespace math
+  {
+    namespace eigen3
+    {
+      inline Eigen::Matrix3d computeCovarianceMatrix(
```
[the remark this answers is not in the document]
scpeters: +1

scpeters (commented):
you've resolved a lot of comments but I don't see the corresponding changes in the code

eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,143 @@
...
+          covariance(0, 0) += pt.x();
+          covariance(0, 1) += pt.y();
+          covariance(0, 2) += pt.z();
+        }
+
+        covariance(1, 0) = covariance(0, 1);
+        covariance(2, 0) = covariance(0, 2);
+        covariance(2, 1) = covariance(1, 2);
+
+        return covariance;
```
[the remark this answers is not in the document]
AmrElsersy: It was the implemntation of [PCL](https://github.com/PointCloudLibrary/pcl/blob/master/common/include/pcl/common/impl/centroid.hpp#L180)

I replaced it with the [implementation](https://github.com/isl-org/Open3D/blob/76c2baf9debd460900f056a9b51e9a80de9c0e64/cpp/open3d/utility/Eigen.cpp#L305) of `Open3D`  and it gave me the same result, let's keep it

AmrElsersy:
> you've resolved a lot of comments but I don't see the corresponding changes in the code

forgot to push  :D

adlarkin (commented):
Overall, this looks good. I'm trying to do a few verifications for testing, but I think that this can be merged as long as I don't find anything strange in the tests I run.

eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,152 @@
...
+  namespace math
+  {
+    namespace eigen3
+    {
+      /// \brief Get covariance matrix from a set of 3d vertices
+      /// https://github.com/isl-org/Open3D/blob/76c2baf9debd460900f056a9b51e9a80de9c0e64/cpp/open3d/utility/Eigen.cpp#L305
+      /// \param[in] _mesh a mesh of 3d vertices
+      /// \return Covariance matrix
+      inline Eigen::Matrix3d covarianceMatrix(
+        const std::vector<math::Vector3d> &_mesh)
```
scpeters, line 45: nit: I would name this `_vertices` instead of `_vertex`

eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,152 @@
...
+        covariance(0, 2) = cumulants(5) - cumulants(0) * cumulants(2);
+        covariance(2, 0) = covariance(0, 2);
+        covariance(1, 2) = cumulants(7) - cumulants(1) * cumulants(2);
+        covariance(2, 1) = covariance(1, 2);
+        return covariance;
+      }
+
+      /// \brief Get the oriented 3d bounding box of a mesh points using PCA
+      /// http://codextechnicanum.blogspot.com/2015/04/find-minimum-oriented-bounding-box-of.html
+      /// \param[in] _mesh a mesh of 3d vertices
```
scpeters, line 80: nit: `_vertices`

eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,152 @@
...
+        covariance(1, 0) = covariance(0, 1);
+        covariance(0, 2) = cumulants(5) - cumulants(0) * cumulants(2);
+        covariance(2, 0) = covariance(0, 2);
+        covariance(1, 2) = cumulants(7) - cumulants(1) * cumulants(2);
+        covariance(2, 1) = covariance(1, 2);
+        return covariance;
+      }
+
+      /// \brief Get the oriented 3d bounding box of a mesh points using PCA
+      /// http://codextechnicanum.blogspot.com/2015/04/find-minimum-oriented-bounding-box-of.html
```
scpeters, line 79: nit: consider rephrasing

`\brief Get the oriented 3d bounding box of a group of vertices using PCA`

eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,152 @@
...
+      /// \param[in] _mesh a mesh of 3d vertices
+      /// \return Oriented 3D box
+      inline ignition::math::OrientedBoxd meshToOrientedBox(
```
scpeters, line 82: nit: `verticesToOrientedBox`

eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,152 @@
...
+        const std::vector<math::Vector3d> &_mesh)
+      {
+        math::Vector3d mean;
+        for (const auto &point : _mesh)
+          mean += point;
+        mean /= static_cast<double>(_mesh.size());
```
scpeters, line 88: should we check the vector size and exit early if the vector is empty to avoid dividing by zero?

scpeters (commented):
I made a few suggestions about documentation and naming, suggesting vertices instead of mesh to avoid confusion with `ignition::common::Mesh`

eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,152 @@
...
+          cumulants(2) += point(2);
+          cumulants(3) += point(0) * point(0);
+          cumulants(4) += point(0) * point(1);
+          cumulants(5) += point(0) * point(2);
+          cumulants(6) += point(1) * point(1);
+          cumulants(7) += point(1) * point(2);
+          cumulants(8) += point(2) * point(2);
+        }
+
+        cumulants /= static_cast<double>(_mesh.size());
```
scpeters: also here: I think we should check the size and exit early if there are no vertices

eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,159 @@
...
+          cumulants(4) += point(0) * point(1);
+          cumulants(5) += point(0) * point(2);
+          cumulants(6) += point(1) * point(1);
+          cumulants(7) += point(1) * point(2);
+          cumulants(8) += point(2) * point(2);
+        }
+
+        Eigen::Matrix3d covariance;
+        if (_vertices.size() == 0)
+          return covariance;
```
scpeters, line 65: this is uninitialized. I would consider returning `Eigen::Matrix3d::Zero()`

osrf-triage:

https://community.gazebosim.org/t/gsoc-2021-machine-learning-extension-to-ignition-gazebo/1070/1

adlarkin:
> I'm trying to do a few verifications for testing

I went ahead and ran the [example set of points in the unit test](https://github.com/ignitionrobotics/ign-math/blob/377623315f5396bf9f51183e84fa31699e259185/eigen3/src/Util_TEST.cc#L28-L61) through [Open3D's `CreateFromPoints` method](https://github.com/isl-org/Open3D/blob/76c2baf9debd460900f056a9b51e9a80de9c0e64/cpp/open3d/geometry/BoundingVolume.h#L123-L129) for comparison. Here are the results I got (link to the code can be found [here](https://gist.github.com/adlarkin/b02bdfa9fd52ea63acb94ee38f164fb8)):

```
Box center is: 
0.423213
0.507995
 1.83459

Box extent (size) is: 
6.53071
4.12782
3.25527

roll: -2.01634
pitch: 0.0058244
yaw: 1.1554
```

It looks like the box's center and size are similar enough between `Open3D` and this PR, but the rotation seems like it could be an issue. Of course, this is assuming that `Open3D` should be considered a "source of truth", but I just wanted to mention it here in case this deserves further investigation. The code I linked shows how I compute roll, pitch and yaw from the rotation matrix that is provided by `Open3D` (hopefully my calculations are correct). cc @scpeters

AmrElsersy:
That is our output for multi-links models (double pendolum and cars) 

image

AmrElsersy:
> @adlarkin
> I thinks that is because the part of calculating the orientation in the 2 algorithms is different
> 
> their code
> 
> ```c++
>     Eigen::Matrix3d R = es.eigenvectors();
>     R.col(0) /= R.col(0).norm();
>     R.col(1) /= R.col(1).norm();
>     R.col(2) /= R.col(2).norm();
> 
>     if (evals(1) > evals(0)) {
>         std::swap(evals(1), evals(0));
>         Eigen::Vector3d tmp = R.col(1);
>         R.col(1) = R.col(0);
>         R.col(0) = tmp;
>     }
>     if (evals(2) > evals(0)) {
>         std::swap(evals(2), evals(0));
>         Eigen::Vector3d tmp = R.col(2);
>         R.col(2) = R.col(0);
>         R.col(0) = tmp;
>     }
>     if (evals(2) > evals(1)) {
>         std::swap(evals(2), evals(1));
>         Eigen::Vector3d tmp = R.col(2);
>         R.col(2) = R.col(1);
>         R.col(1) = tmp;
>     }
> ```

That was the output when I tried that code of Open3D for calculating Rotation instead of the current algorithm
It was bad

ezgif com-gif-maker (2)

eigen3/src/Util_TEST.cc
```diff
@@ -0,0 +1,62 @@
...
+
+#include <gtest/gtest.h>
+
+#include <ignition/math/eigen3/Util.hh>
+
+using namespace ignition;
+
+/////////////////////////////////////////////////
+/// \brief Test the oriented box converted from a set of vertices
+TEST(EigenUtil, verticesToOrientedBox)
```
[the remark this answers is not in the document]
AmrElsersy: mmmm the Simple box approach worked ! and gave me the exact box 

I tried that before but on a rotated box in all directions, and it didn't gave me the same box, so that is why I told you that it will not work

eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,159 @@
...
+{
+  namespace math
+  {
+    namespace eigen3
+    {
+      /// \brief Get covariance matrix from a set of 3d vertices
+      /// https://github.com/isl-org/Open3D/blob/76c2baf9debd460900f056a9b51e9a80de9c0e64/cpp/open3d/utility/Eigen.cpp#L305
+      /// \param[in] _vertices a vector of 3d vertices
+      /// \return Covariance matrix
+      inline Eigen::Matrix3d covarianceMatrix(
```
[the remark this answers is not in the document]
AmrElsersy: Ok

eigen3/include/ignition/math/eigen3/Util.hh
```diff
@@ -0,0 +1,161 @@
...
+          -1.0f * (projectionTransform.block<3, 3>(0, 0) * centroid);
+
+        Eigen::Vector3d minPoint(INF_I32, INF_I32, INF_I32);
+        Eigen::Vector3d maxPoint(-INF_I32, -INF_I32, -INF_I32);
+
+        // Get the minimum and maximum points of the transformed cloud.
+        for (const auto &point : _vertices)
+        {
+          Eigen::Vector4d pt(0, 0, 0, 1);
+          pt.head<3>() = convert(point);
```
adlarkin, line 129: Is this one of the `math::eigen3::convert` methods? If so, we can add the `math::eigen3` namespace so that it's easy for code readers to know which function is being used (adding this namespace would also help avoid potential collisions since `convert` is a generic method name).

eigen3/src/Util_TEST.cc
```diff
@@ -0,0 +1,153 @@
...
+  EXPECT_DOUBLE_EQ(size.Y(), 0);
+  EXPECT_DOUBLE_EQ(size.Z(), 0);
+
+  EXPECT_DOUBLE_EQ(position.X(), 0);
+  EXPECT_DOUBLE_EQ(position.Y(), 0);
+  EXPECT_DOUBLE_EQ(position.Z(), 0);
+
+  EXPECT_DOUBLE_EQ(rotation.Roll(), 0);
+  EXPECT_DOUBLE_EQ(rotation.Pitch(),0);
+  EXPECT_DOUBLE_EQ(rotation.Yaw(),  0);
```
adlarkin, line 87:
```suggestion
  EXPECT_DOUBLE_EQ(rotation.Pitch(), 0);
  EXPECT_DOUBLE_EQ(rotation.Yaw(), 0);
```
nit: fix whitespace

eigen3/src/Util_TEST.cc
```diff
@@ -0,0 +1,153 @@
...
+  EXPECT_NEAR(position.Y(), 0, error);
+  EXPECT_NEAR(position.Z(), 0, error);
+
+  EXPECT_NEAR(rotation.Roll(), 0, error);
+  EXPECT_NEAR(rotation.Pitch(), 0, error);
+  EXPECT_NEAR(rotation.Yaw(), 0, error);
+}
+
+/////////////////////////////////////////////////
+TEST(EigenUtil, covarianceTest)
```
adlarkin, line 127: Can you add one more test for `math::eigen3::covarianceMatrix` where an empty set of vertices is passed in? We should receive an identity matrix in this case.

adlarkin (commented):
Just a few small comments, but I think this is almost ready to merge!

scpeters:
have we considered specifying the orientation of the bounding box and then simply expanding its dimensions to include all the passed vertices? this would avoid the weird box rotations at the cost of generating non-minimal boxes

adlarkin:
> have we considered specifying the orientation of the bounding box and then simply expanding its dimensions to include all the passed vertices? this would avoid the weird box rotations at the cost of generating non-minimal boxes

I think the challenge is that in some scenarios, we have the vertices of an object, but not the orientation of the object. For example, in https://github.com/ignitionrobotics/ign-rendering/pull/334, I believe that we have all of the vertices from the bounding boxes provided by Ogre, but I don't believe that we have a way of getting the "combined orientation" from all of the rendering visuals (I could be wrong about this).

However, that being said, I think that it would be good to provide users with the option to pass in an orientation to `verticesToOrientedBox`. Perhaps we can add a input parameter that's a pointer to an orientation: if the pointer is null, we ignore it and run the algorithm as-is, but if it isn't null, then we take the orientation into account. Does that sound reasonable?

adlarkin (commented):
Code looks good! I'm going to wait to see what @scpeters thinks about https://github.com/ignitionrobotics/ign-math/pull/224#issuecomment-909325297 before merging this.

scpeters:
> However, that being said, I think that it would be good to provide users with the option to pass in an orientation to `verticesToOrientedBox`. Perhaps we can add a input parameter that's a pointer to an orientation: if the pointer is null, we ignore it and run the algorithm as-is, but if it isn't null, then we take the orientation into account. Does that sound reasonable?

I would just add a separate API that accepts a vector of vertices as well as a `Quaternion`, but this PR is fine without it. We could add it later if desired

adlarkin (approved):

PR #229, opened 2021-08-25
Adds python interface to PID and SemanticVersion.

LolaSegura:
Signed-off-by: LolaSegura <lsegura@ekumenlabs.com>
# 🎉 New feature
Related to #101 #210

## Summary
Adds Python interface for two math classes: PID, SemanticVersion. For each class a python test has been created.

### PID
* A copy constructor was added to the python class. The PID C++ class didn't have one defined, and, because on python the `operator=` is not supported, there was no way to copy two variables of type PID. 
* The `Update(const double _error, const std::chrono::duration<double> &_dt)` method received a variable of type `std::chrono::duration`. To give support to this method in python a new  update function was defined using the `%extend` command. This new update method receives a variable of type double `double Update(const double error, double dt)`.
* The `void Errors(double &_pe, double &_ie, double &_de) const;`  method changed the values of the arguments inside it. For native types swig has a way to handle this  using the `%apply <type> *OUTPUT {}` directive. So in python this method can be used as `[pe, ie, de] = pid.errors()`
* The `%rename` command is failing on adding and under case when the variable name is conformed by one letter and a world, for example, `PGain()` was being renamed as `pgain()` instead of `p_gain()`. So this cases where contemplated using an specific `%rename` tag.

## SemanticVersion
* The `std_string.i` module was added in order to work with `std::string`.

## Checklist
- [x] Added tests

### DIFF
--- a/src/python/CMakeLists.txt
+++ b/src/python/CMakeLists.txt
@@ -17,7 +17,9 @@ if (SWIG_FOUND)
   set(swig_files
     Angle
     GaussMarkovProcess
+    PID
     Rand
+    SemanticVersion
     Vector2
     Vector3
     Vector4)
@@ -70,8 +72,10 @@ if (PYTHONLIBS_FOUND)
     set(python_tests
       Angle_TEST
       GaussMarkovProcess_TEST
+      PID_TEST
       python_TEST
       Rand_TEST
+      SemanticVersion_TEST
       Vector2_TEST
       Vector3_TEST
       Vector4_TEST
--- a/src/python/PID.i
+++ b/src/python/PID.i
@@ -1,0 +1,89 @@
+/*
+ * Copyright (C) 2021 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


chapulina (approved):
LGTM!

> The Update(const double _error, const std::chrono::duration<double> &_dt) method received a variable of type std::chrono::duration. To give support to this method in python a new update function was defined using the %extend command. This new update method receives a variable of type double double Update(const double error, double dt).

I think this is ok for now, but maybe in the future we should look into using python time to be more expressive?

francocipollone:
> LGTM!
> 
> > The Update(const double _error, const std::chrono::duration &_dt) method received a variable of type std::chrono::duration. To give support to this method in python a new update function was defined using the %extend command. This new update method receives a variable of type double double Update(const double error, double dt).
> 
> I think this is ok for now, but maybe in the future we should look into using python time to be more expressive?

Agree. I think that we should provide a conversion from std::chrono to python time. Sadly, swig doesn't provide bindings as it does for `std::string`(std_string.p) or `std::map` (std_map.i) so probably creating a `std_chrono.i` might be useful to have, keeping in mind that there is this other class, `Stopwatch`, that also uses `std::chrono` extensively.

PR #230, opened 2021-08-30
Adds python interface to Filter, MovingWindowFilter, RotationSpline.

LolaSegura:
# 🎉 New feature
Goes on top of #221
Related to #101 #210

## Summary
Adds Python interface for three math classes: Filter, MovingWindowFilter and RotationSPline. For each class a python test has been created.

### Filter
The `Filter.hh` file contains  various template classes that inherit from a `Filter` class, which is a template class itself. To create binding for this classes an instance of `Filter<type>` had to be declared previously. 
There were some classes that inherit from an instance of another class. For example `class OnePoleQuaternion : public OnePole<math::Quaterniond>`. This wasn't being interpreted by swig as expected. The inherited methods that were not  override, were not being created in the binding. The solution we implemented for this was define these classes inside the interface file without the inheritance and adding the methods that were not being created. A similar approach was taken in the `SignalStats` class (#220 ).

## Checklist
- [x] Added tests


francocipollone:
This PR is on hold until #221 is merged.

francocipollone (approved):
LGTM

francocipollone:
I'll solve the conflicts after #221 is merged.

src/python/MovingWindowFilter.i
```diff
@@ -0,0 +1,47 @@
...
+ *
+*/
+%module movingwindowfilter
+%{
+#include "ignition/math/Export.hh"
+#include "ignition/math/MovingWindowFilter.hh"
+#include "ignition/math/Vector3.hh"
+%}
+
+%import "std_vector.i"
```
azeey, line 24: Removing this line fixes the compilation error for me. Maybe it's a known issue, but I had to `touch python.i` for it to recompile this file.

francocipollone: Let me take a look. It seemed to be working before merging this branch with ign-math6.

francocipollone: You were right. That import was causing issues because of the syntaxis: it should be `include` instead of `import`

The truth is that that import wasn't necessary given that `std::vector` isn't being used in the API, so that import statement wasn't needed at all. I pushed the fix

scpeters (approved):

PR #234, opened 2021-08-31
Adds python interface to MaterialType and Material.

LolaSegura:
# 🎉 New feature

Related to #101 #210

## Summary
Adds Python interface for three math classes: MaterialType, Material. For each Material a python test has been created.

### MaterialType
This file contains an `enum class` declaration. Swig supports strongly typed enumerations, so binding this wasn't a problem. But the result was a series of variables contained in the module `ignition.math` defined as `MaterialType_MATERIAL`. A work around is needed to be implemented in order to follow the pep-8 naming convention.

### Material
**The test is not passing.**
* This class contains a method `Predefined()` that returns a reference to an `std::map<MaterialType, Material>`.  In  `SignalStats`a map was used as well, so a similar approach was taken for this case. There are two main differences in both of this classes:
   * In `SignalStats` the map is returned as a copy and 
   * the types of the map are `std::string` and `double`. 
  
* In the case of `Material` the method is not being wrapped correctly. Instead of being part of the `Material` class, is defined outside of it, and when it is used in the test as follow:

      mats = ignition.math.Material.predefined() 

   it arises an error:
   ```
   swig/python detected a memory leak of type 'unknown', no destructor found.
      TypeError: unhashable type: 'SwigPyObject'
      The above exception was the direct cause of the following exception:

        File "/ws/src/ign-math/src/python/Material_TEST.py", line 22, in test_init
           mats = ignition.math.Material.predefined()
       SystemError: <built-in function Material_predefined> returned a result with an error set 


## Checklist
- [x] Added tests

### DIFF
--- a/src/python/CMakeLists.txt
+++ b/src/python/CMakeLists.txt
@@ -13,14 +13,6 @@ if (SWIG_FOUND)
 
   include_directories(${PROJECT_SOURCE_DIR}/include)
   include_directories(${PYTHON_INCLUDE_PATH})
-
-  set(swig_files
-    Angle
-    GaussMarkovProcess
-    Rand
-    Vector2
-    Vector3
-    Vector4)
 endif()
 
 #################################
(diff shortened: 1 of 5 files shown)


francocipollone:
Thanks @ahcorde for taking over the PR and completing it! :rocket:

PR #235, opened 2021-09-03
🌐 Spherical coordinates: bug fix, docs and sanity checks

chapulina:
# 🦟 Bug fix

Fixes https://github.com/osrf/gazebo/issues/2022
Part of https://github.com/ignitionrobotics/ign-gazebo/issues/981

## Summary

I believe that https://github.com/osrf/gazebo/issues/2022 is only an issue with the `*FromLocal*` functions. While the heading rotation from `GLOBAL` to `LOCAL` seems to be [correct](https://en.wikipedia.org/wiki/Rotation_matrix), the rotation from `LOCAL` to `GLOBAL` was not the inverse of that, which resulted in incoherent behaviour.

In order to preserve behaviour on version 6, which is being used by a lot of people who compensate for the bug, I introduced a new frame, `LOCAL2`, which can be used to get the correct calculation.

I also added several tests with sanity checks (i.e. higher latitude means further North...). 

It's my first time playing with this class, so I'd appreciate it if people paid close attention to the math and test cases, in case I'm misunderstanding something.

## Checklist
- [x] Added tests

### DIFF
--- a/include/ignition/math/SphericalCoordinates.hh
+++ b/include/ignition/math/SphericalCoordinates.hh
@@ -34,7 +34,6 @@ namespace ignition
     //
     class SphericalCoordinatesPrivate;
 
-    /// \class SphericalCoordinates SphericalCoordinates.hh commmon/common.hh
     /// \brief Convert spherical coordinates for planetary surfaces.
     class IGNITION_MATH_VISIBLE SphericalCoordinates
     {
@@ -61,7 +60,12 @@ namespace ignition
                 GLOBAL = 3,
 
                 /// \brief Heading-adjusted tangent plane (X, Y, Z)
-                LOCAL = 4
+                /// This has kept a bug for backwards compatibility, use
+                /// LOCAL_FIXED for the correct behaviour.
+                LOCAL = 4,
+
+                /// \brief Heading-adjusted tangent plane (X, Y, Z)
+                LOCAL_FIXED = 5
               };
 
       /// \brief Constructor.
@@ -91,7 +95,14 @@ namespace ignition
       public: ~SphericalCoordinates();
 
       /// \brief Convert a Cartesian position vector to geodetic coordinates.
-      /// \param[in] _xyz Cartesian position vector in the world frame.
+      /// This performs a `PositionTransform` from LOCAL to SPHERICAL.
+      ///
+      /// There's a known bug with this computation that can't be fixed on
+      /// version 6 to avoid behaviour changes. Directly call
+      /// `PositionTransform(_xyz, LOCAL_FIXED, SPHERICAL)` for correct results.
+      ///
+      /// \param[in] _xyz Cartesian position vector in the heading-adjusted
+      /// world frame.
       /// \return Cooordinates: geodetic latitude (deg), longitude (deg),
       ///         altitude above sea level (m).
       public: ignition::math::Vector3d SphericalFromLocalPosition(
@@ -99,7 +110,14 @@ namespace ignition
 
       /// \brief Convert a Cartesian velocity vector in the local frame
       ///        to a global Cartesian frame with components East, North, Up.
-      /// \param[in] _xyz Cartesian velocity vector in the world frame.
+      /// This is a wrapper around `VelocityTransform(_xyz, LOCAL, GLOBAL)`
+      ///
+      /// There's a known bug with this computation that can't be fixed on
+      /// version 6 to avoid behaviour changes. Directly call
+      /// `VelocityTransform(_xyz, LOCAL_FIXED, GLOBAL)` for correct results.
+      ///
+      /// \param[in] _xyz Cartesian velocity vector in the heading-adjusted
+      /// world frame.
       /// \return Rotated vector with components (x,y,z): (East, North, Up).
       public: ignition::math::Vector3d GlobalFromLocalVelocity(
                   const ignition::math::Vector3d &_xyz) const;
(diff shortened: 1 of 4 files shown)


src/SphericalCoordinates_TEST.cc
```diff
@@ -383,3 +432,161 @@ TEST(SphericalCoordinatesTest, AssignmentOp)
...
+
+  // Increase longitude == go East == go +X (and a bit -Y)
+  {
+    auto xyz = sc.LocalFromSphericalPosition(
+        {lat.Degree(), lon.Degree() + 1.0, elev});
+    EXPECT_LT(xyzOrigin.X(), xyz.X());
+    EXPECT_GT(xyzOrigin.Y(), xyz.Y());
+  }
+
+  // Decrease longitude == go West == go -X (and a bit -Y)
```
scpeters, line 482: it seems weird that a change to longitude in either direction causes Y to decrease slightly. I would have thought it would be linearized

src/SphericalCoordinates_TEST.cc
```diff
@@ -383,3 +432,161 @@ TEST(SphericalCoordinatesTest, AssignmentOp)
...
+}
+
+//////////////////////////////////////////////////
+TEST(SphericalCoordinatesTest, WithHeading)
+{
+  // Heading 90 deg: X == North, Y == West , Z == Up
+  auto st = math::SphericalCoordinates::EARTH_WGS84;
+  math::Angle lat(IGN_DTOR(-22.9));
+  math::Angle lon(IGN_DTOR(-43.2));
+  math::Angle heading(0.0);
```
scpeters, line 521: set this to 90 degrees

src/SphericalCoordinates_TEST.cc
```diff
@@ -383,3 +432,161 @@ TEST(SphericalCoordinatesTest, AssignmentOp)
...
+  double elev = 0;
+  math::SphericalCoordinates sc(st, lat, lon, elev, IGN_DTOR(90));
```
scpeters, line 523: use `heading` instead of `IGN_DTOR(90)`

src/SphericalCoordinates_TEST.cc
chapulina: I was also surprised to see this, I assumed it was an artifact of the projection and didn't look further into it.

chapulina: Ok, so I spent some time looking into this and I think I understand what's happening. This image should help:

image

Note that the ENU plane is tangent to the ellipsoid.

What this test is doing is changing lambda while keeping phi fixed. As you do that, the projection of the point onto the plane will only move on a straight line if you're exactly on the equator. If you're on the northern hemisphere, it goes further North (+Y) and if you're in the southern hemisphere, like on the test, it goes further South (-Y).

Here's a handy way of visualizing it, see how the projection of the blue line is not linear:

image

https://demonstrations.wolfram.com/TangentPlaneToASphere/

src/SphericalCoordinates_TEST.cc
```diff
@@ -383,3 +432,210 @@ TEST(SphericalCoordinatesTest, AssignmentOp)
...
+
+  // Decrease longitude == go West == go -X
+  // Also move a bit -Y because this is the Southern Hemisphere
+  {
+    auto xyz = sc.LocalFromSphericalPosition(
+        {lat.Degree(), lon.Degree() - 1.0, elev});
+    EXPECT_GT(xyzOrigin.X(), xyz.X());
+    EXPECT_GT(xyzOrigin.Y(), xyz.Y());
+  }
+
```
mjcarroll, line 491: Tests for increasing/decreasing altitude?  Should be consistent because of right hand rule, but a reasonable sanity check.

src/SphericalCoordinates_TEST.cc
```diff
@@ -383,3 +432,210 @@ TEST(SphericalCoordinatesTest, AssignmentOp)
...
+  // Check how global and local velocities are connected
+
+  // Velocity in +X (East), +Y (North), -X (West), -Y (South)
```
mjcarroll, line 494: Also check velocity in +z/-z

mjcarroll (commented):
Maybe consider adding +/- altitude tests as well for completeness.

chapulina:
> Maybe consider adding +/- altitude tests as well for completeness.

Done in a8933ca

scpeters (approved):

include/ignition/math/SphericalCoordinates.hh
```diff
@@ -91,15 +95,31 @@ namespace ignition
       public: ~SphericalCoordinates();
 
       /// \brief Convert a Cartesian position vector to geodetic coordinates.
-      /// \param[in] _xyz Cartesian position vector in the world frame.
+      /// This performs a `PositionTransform` from LOCAL to SPHERICAL.
+      ///
+      /// There's a known bug with this computation that can't be fixed on
+      /// version 6 to avoid behaviour changes. Directly call
```
peci1, line 101: @scpeters The comments indicate that the behavior was only intended to be kept for gz-math6. However, it slipped through to gz-math7. Is there some will to straighten this up in 8, or will we live with this forever?

scpeters: sure, we can fix it in gz-math8

can you open an issue or pull request to `main`?

peci1: Good, I'll try the PR.

I've noticed another discrepancy in the interface: SphericalFromLocalPosition returns spherical coordinates in degrees, while PositionTransform uses radians. Shouldn't this be also united to using degrees everywhere? Or was there a reason for keeping it this way?

scpeters: I think consistency would be preferred; it was probably oversight in the first place that caused them to diverge. A `std::vector<math::Angle>` type would handle the units more clearly but it a little more overhead than `Vector3d`

peci1: Okay, I've sent the PRs: #596, #597.

PR #236, opened 2021-09-06
Add options to install python/ruby in system standard paths

j-rivero:
# 🎉 New feature

## Summary

Introduce two new options: `USE_SYSTEM_PATHS_FOR_PYTHON_INSTALLATION` and `USE_DIST_PACKAGES_FOR_PYTHON` to support python installation in the standard directories. The local installations can go in `<path>/site-packages` (i.e /usr/lib/python3/site-packages) but for packaging Debian is recommending to use `<path>/dist-packages`.

The best way to implement this is to use the `find_package(Python3)` but is only available from 3.12 on so I've implemented a workaround invoking 

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,10 @@ ign_configure_project(VERSION_SUFFIX)
 # Set project-specific options
 #============================================================================
 
+option(USE_SYSTEM_PATHS_FOR_RUBY_INSTALLATION
+      "Install ruby modules in standard system paths in the system"
+      OFF)
+
 option(USE_SYSTEM_PATHS_FOR_PYTHON_INSTALLATION
       "Install python modules in standard system paths in the system"
       OFF)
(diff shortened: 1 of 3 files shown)


chapulina:
This is being tested in https://github.com/ignition-release/ign-math6-release/pull/6

chapulina (approved):
LGTM

I noticed that SWIG is not being found on macOS and Windows CI even without this PR. So note that these paths haven't really been tested on those platforms. Not this PR's fault though.

scpeters:
> LGTM
> 
> I noticed that SWIG is not being found on macOS and Windows CI even without this PR. So note that these paths haven't really been tested on those platforms. Not this PR's fault though.

we can enable SWIG by adding it as a dependency to the [ignition-math6](https://github.com/osrf/homebrew-simulation/blob/master/Formula/ignition-math6.rb) formula. I just tested it locally and there are some extra compiler warnings and test failures related to random number generation in Ruby.

chapulina:
> we can enable SWIG 

Ticketed https://github.com/ignition-tooling/release-tools/issues/508

PR #239, opened 2021-09-16
Use python executable variable instead of hardcode python3

j-rivero:
This should prepare the software for https://github.com/ignition-tooling/release-tools/issues/508. Note that CI won't run the python/ruby test on Mac/Windows until #508 is ready.

### DIFF
--- a/src/python/CMakeLists.txt
+++ b/src/python/CMakeLists.txt
@@ -110,7 +110,7 @@ if (PYTHONLIBS_FOUND)
 
     foreach (test ${python_tests})
       add_test(NAME ${test}.py COMMAND
-        python3 ${CMAKE_SOURCE_DIR}/src/python/${test}.py)
+        "${Python3_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/src/python/${test}.py")
 
       set(_env_vars)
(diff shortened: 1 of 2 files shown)


chapulina (approved):

src/python/CMakeLists.txt
```diff
@@ -110,7 +110,7 @@ if (PYTHONLIBS_FOUND)
 
     foreach (test ${python_tests})
       add_test(NAME ${test}.py COMMAND
-        python3 ${CMAKE_SOURCE_DIR}/src/python/${test}.py)
+        "${Python3_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/src/python/${test}.py")
```
scpeters, line 113: which cmake call is providing this variable?

j-rivero: `find_package(Python3)` here https://github.com/ignitionrobotics/ign-math/pull/239/files#diff-43c9042647d14a73e4061f77bba19ee368fd059a7e0c6086063f7646e2b2aabbR69. Umm, that call is conditionally scoped, I thought it was used in the root CMakeLists.txt but it is the old `find_package(PythonLibs)`. We don't have `find_package(Python3)` in Bionic, I changed the code to use old `PYTHON_EXECUTABLE` variable.

scpeters (approved):
works on macOS for me now

PR #240, opened 2021-09-16
Suggestions to #219

chapulina:
Some suggestions while reviewing the PR:

* Added more tests for `Plane::Intersect` (but I think we should add more once the plane's size is taken into consideration)
* Extended `Line3::Distance` to consider the line's limits (see `Vector3::DistToLine` for an existing function that works with infinite lines, also added more tests there)
* Made some updates to `Sphere::VolumeBelow` to handle vertical planes, and I also believe that the function was originally returning the volume above, see the added tests.
* Started adding tests for `Sphere::CenterOfVolumeBelow`
* Small style updates

I'm still reviewing the PR, I haven't looked into `Box` yet.

### DIFF
--- a/include/ignition/math/detail/Sphere.hh
+++ b/include/ignition/math/detail/Sphere.hh
@@ -103,18 +103,24 @@ T Sphere<T>::VolumeBelow(const Planed &_plane) const
   // get nearest point to center on plane
   auto dist = _plane.Distance(Vector3d(0, 0, 0));
 
-  if(dist < -r)
+  if (dist < -r)
   {
     // sphere is completely below plane
     return Volume();
   }
-  else if(dist > r)
+  else if (dist > r)
   {
     // sphere is completely above plane
-    return 0;
+    return 0.0;
   }
 
-  auto h = r + dist;
+  // Vertical plane
+  if (_plane.Normal().Z() < 1e-6)
+  {
+    return 0.0;
+  }
+
+  auto h = r - dist;
   return IGN_PI * h * h * (3 * r - h) / 3;
 }
 
--- a/include/ignition/math/Line3.hh
+++ b/include/ignition/math/Line3.hh
@@ -230,12 +230,29 @@ namespace ignition
       }
 
       /// \brief Calculate shortest distance between line and point
-      /// \param[in] _pt point which we are measuring distance to.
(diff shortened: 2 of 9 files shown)


include/ignition/math/detail/Sphere.hh
```diff
@@ -103,18 +103,24 @@ T Sphere<T>::VolumeBelow(const Planed &_plane) const
...
+  else if (dist > r)
   {
     // sphere is completely above plane
-    return 0;
+    return 0.0;
   }
 
-  auto h = r + dist;
+  // Vertical plane
+  if (_plane.Normal().Z() < 1e-6)
```
arjo129, line 118: Why should this be the case? I consider "below" to be defined by the normal. It is possible that the sphere is rotated when surfacing.

arjo129 (commented):
Thanks for the :bug: fixes! Most of the changes look good to me. However, I don't agree with the sphere and the vertical plane. I guess I need to be more articulate in my documentation but volume below should return the Volume "below" the plane where "up" refers to the normal of the line. The rationale for this is in future revisions of the buoyancy plugin, I'd like to be rotate the object/plane according to its pose.

include/ignition/math/detail/Sphere.hh
chapulina: Ah ok, I assumed the opposite, and even updated the docs :sweat_smile: 

https://github.com/ignitionrobotics/ign-math/blob/f398782fd650547d2d1be9693a9e7569bbe38e3a/include/ignition/math/Sphere.hh#L97

In that case, I think the implementation needs to be fixed to reflect that. See this test that I added, the normal doesn't seem to affect the result:

https://github.com/ignitionrobotics/ign-math/blob/f398782fd650547d2d1be9693a9e7569bbe38e3a/src/Sphere_TEST.cc#L141-L149

chapulina: Reading that doc update now, I think it's badly worded :grimacing: I meant to say that the normal's direction is not taken into account.

arjo129: Yep will need to fix that. I can merge these changes in and then work on a fix for it.

arjo129 (approved):

PR #243, opened 2021-09-21
More suggestions to #219

chapulina:
* Style updates
* Use `set` instead of `vector` to avoid duplicates and return a predictable order (sorted). Without this, sometimes `Intersections` may return duplicate points.
* Use `Precision` for the functions in `Sphere`
* Moved some duplicate logic to a new `Box::VerticesBelow` function
* Added more test cases


arjo129 (approved):
Windows seems to be computing things differently will investigate.

Issue #269, opened 2021-10-15
Unsafe Constant Initialization


mahiuchun:
In `ignition::math::Color` the color constants are defined like `public: static const Color Cyan;`. While the resulting interface is nice to use, it could cause static initialization order fiasco as C++ makes no guarantee that global variables in this form is going to be initialized in usage order. In other words, another global variable could use `Color::Cyan` and be initialized first.

A simple fix is to convert to a function form like the following:
`public: const Color& Cyan() { static auto *c = new Color(0, 1, 1, 1); return *c; }`

Another solution is more involved but it could retain the ability to refer to constants without parentheses.
1. Add a constexpr constructor to the class.
2. In another class, say, ColorConstants, use `public: static constexpr Color Cyan = {0, 1, 1, 1};`

Both solutions represent a API/ABI change, so I would like to hear some feedback first before sending out PRs.

The Vector classes might be affected by this too.

jwnimmer-tri:
> The Vector classes might be affected by this too.

Yes, I believe that `Vector3<T>::UnitX` etc. could all affected, at least in theory.  (Some compilers will optimize away the problematic portions, in practice.)

The full list of static const globals on the default branch seems to be: Angle, Color, Material, Matrix3, Matrix4, Pose3, Quaternion, Vector2, Vector3, Vector4.

Note that `Material::Predefined()` is particularly problematic, because it returns to the user a reference to `std::map`, which always implies a heap allocation and thus will always be leaked anytime the library is loaded and then unloaded.  Similarly for `kMaterialData` in `src/MaterialType.hh`.  Even if those collection types are fixed to be heapless, the `class Material` has a name which is required to always be on the heap.

Also note that both construction _and destruction_ order are in play here.  As library code, I'd argue that ign-math should have no global constructors nor global destructors registered at all.

For cross-reference, see also https://github.com/ignitionrobotics/sdformat/issues/552.

> A simple fix is ...

[...]

PR #280, opened 2021-11-05
Vector2, 3 and 4 with pybind11

ahcorde:
Signed-off-by: ahcorde <ahcorde@gmail.com>

# 🎉 New feature

## Summary

This is a draft PR tothat create wrappers for Vector2, 3 and 4 with pybind11

## Checklist
- [ ] Signed all commits for DCO
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge**

PR #283, opened 2021-11-22
Use constexpr for simple static constants

jwnimmer-tri:
# 🦟 Bug fix

A portion of #269.

## Summary

Use constexpr for simple static constants, to avoid the C++ global static construction and destruction order fiascos.

At the moment, only Color has been refactored with this pattern while we discuss the implications.

This is an alternative approach vs #286.   In thisversion of the fix, we remain API compatible (but still break ABI).  In the #286 version of the fix, we have an API+ABI break (with a deprecation transition), and thus will require users to refactor their code. 

~This pull request is currently targeted at `ign-math6`, but this is an ABI-breaking change (not API-breaking), so I think this should be targeted to `main` or `ign-math7` instead, but those branches do not seem to exist yet.~

## Checklist
- [x] ~Added tests (n/a)~
- [x] ~Updated documentation (as needed) (n/a)~
- [x] ~Updated migration guide (as needed) (n/a)~


scpeters:
> This pull request is currently targeted at `ign-math6`, but this is an ABI-breaking change (not API-breaking), so I think this should be targeted to `main` or `ign-math7` instead, but those branches do not seem to exist yet.

I believe we have a `main` branch:

* https://github.com/ignitionrobotics/ign-math/tree/main

we probably have too many other branches so it doesn't show up in the list unless you start typing its name. Yes, if this breaks ABI, then it should target `main`

scpeters:
there seems to be an issue with our windows CI as well `error C2491: 'ignition::math::v6::Color::White': definition of dllimport static data member not allowed`

~~~
C:\Jenkins\workspace\ignition_math-ci-pr_any-windows7-amd64\ws\ign-math\include\ignition/math/Color.hh(351): error C2491: 'ignition::math::v6::Color::White': definition of dllimport static data member not allowed
C:\Jenkins\workspace\ignition_math-ci-pr_any-windows7-amd64\ws\ign-math\include\ignition/math/Color.hh(352): error C2491: 'ignition::math::v6::Color::Black': definition of dllimport static data member not allowed
C:\Jenkins\workspace\ignition_math-ci-pr_any-windows7-amd64\ws\ign-math\include\ignition/math/Color.hh(353): error C2491: 'ignition::math::v6::Color::Red': definition of dllimport static data member not allowed
C:\Jenkins\workspace\ignition_math-ci-pr_any-windows7-amd64\ws\ign-math\include\ignition/math/Color.hh(354): error C2491: 'ignition::math::v6::Color::Green': definition of dllimport static data member not allowed
C:\Jenkins\workspace\ignition_math-ci-pr_any-windows7-amd64\ws\ign-math\include\ignition/math/Color.hh(355): error C2491: 'ignition::math::v6::Color::Blue': definition of dllimport static data member not allowed
C:\Jenkins\workspace\ignition_math-ci-pr_any-windows7-amd64\ws\ign-math\include\ignition/math/Color.hh(356): error C2491: 'ignition::math::v6::Color::Yellow': definition of dllimport static data member not allowed
C:\Jenkins\workspace\ignition_math-ci-pr_any-windows7-amd64\ws\ign-math\include\ignition/math/Color.hh(357): error C2491: 'ignition::math::v6::Color::Magenta': definition of dllimport static data member not allowed
C:\Jenkins\workspace\ignition_math-ci-pr_any-windows7-amd64\ws\ign-math\include\ignition/math/Color.hh(358): error C2491: 'ignition::math::v6::Color::Cyan': definition of dllimport static data member not allowed
~~~

jwnimmer-tri:
> I believe we have a main branch ...

Oops!  You're right, of course.  I've re-targeted the pull request now.

> there seems to be an issue with our windows CI ...

Yup, I'll work on that error using godbolt.  (I don't remember the dllimport/dllexport rules for constexpr off the top of my head.)  MSVC support is definitely the biggest risk with this approach.  I'll see if I can get it working.

jwnimmer-tri:
> > there seems to be an issue with our windows CI ...
>
> MSVC support is definitely the biggest risk with this approach.

I spent a few hours and couldn't figure out how to get MSVC to obey `inline` (as implied by `constexpr`) on a storage definition, when mixed with dllexport.  I can't tell if its a bug in MSVC or my understanding of how dllexport is supposed to work.  In any case, playing compiler games is probably not the path we want to take here, so I abandoned that approach.

Instead, I switched this implementation to use const references for the globals, but initialize them using references to constexpr storage.  I believe this will be fiasco-safe.

I'm seeking input from project maintainers on whether this (#283) or else #286 is your preferred tactic to resolve the fiasco problem.  Once I have that, I'll close the dispreferred PR, and update the preferred PR to cover all of the relevant types (Vector2, Vector3, etc.) that use the same pattern.

azeey:
Per my [comment](https://github.com/ignitionrobotics/ign-math/issues/269#issuecomment-991230857) in the issue, we prefer #283 over #286 as it maintains API stability.

jwnimmer-tri:
Okay, I've started by rebasing this onto the latest `main` and finishing up the port for Color.

I'll work on porting more classes to this pattern in a future push.

jwnimmer-tri:
This should be ready for feedback / review now.

mahiuchun:
Hi, thanks for working on this. I wonder would the newly added references suffer from initialization order fiasco as well.

jwnimmer-tri:
> I wonder would the newly added references suffer from initialization order fiasco as well.

No, I don't think that reference initialization generates any global constructors or destructors.

scpeters (approved):
this LGTM, thanks for your excellent work here!

my one concern is that the `ign-math6` branch is very out of date and some new static `NaN` constants were added to the `Vector` classes in https://github.com/ignitionrobotics/ign-math/pull/222 (released in 6.9.0). I've started merging forward incrementally, starting from the 6.8.0 tag in #305.

Do you want to wait until after we have fully merged forward from `ign-math6` and resolve conflicts in this branch or merge this now and fix the `NaN` implementation when merging forward?

jwnimmer-tri:
I'm content to do it either way, whatever is easiest overall.  My goal is to get this fix into Drake's build eventually, but that means a numbered release tag here and porting libsdformat to use it, so nothing here is time-sensitive for me.

scpeters:
> I'm content to do it either way, whatever is easiest overall. My goal is to get this fix into Drake's build eventually, but that means a numbered release tag here and porting libsdformat to use it, so nothing here is time-sensitive for me.

thanks for your flexibility. I just finished merging #222 forward to `main`, so the promised conflicts are now ready to be resolved

scpeters:
> > I'm content to do it either way, whatever is easiest overall. My goal is to get this fix into Drake's build eventually, but that means a numbered release tag here and porting libsdformat to use it, so nothing here is time-sensitive for me.
> 
> thanks for your flexibility. I just finished merging #222 forward to `main`, so the promised conflicts are now ready to be resolved

actually only Vector3 seems to have conflicts, though Vector2 and Vector4 will need updates as well

jwnimmer-tri:
Done.

jwnimmer-tri:
If someone can figure out (and push a fix for) the non-C++ test failures, I'd appreciate it.  Looks like maybe the swig file needs updating?

scpeters:
> If someone can figure out (and push a fix for) the non-C++ test failures, I'd appreciate it. Looks like maybe the swig file needs updating?

sure; I will take a look and possibly ask @ahcorde for help if I get stuck

ahcorde:
According with DGB code is failing here:

```cpp
      /// \brief Corrects any nan values
      public: inline void Correct()
      {
        // std::isfinite works with floating point values,
        // need to explicit cast to avoid ambiguity in vc++.
        if (!std::isfinite(static_cast<double>(this->data[0])))
          this->data[0] = 0;
        if (!std::isfinite(static_cast<double>(this->data[1])))
          this->data[1] = 0;
      }
```

scpeters:
@ahcorde I've cherry-picked #280 and #285 to `main` and resolved conflicts. Can you take a look if there are test failures?

azeey:
The failing tests are attempting to mutate a reference to `Vector*d::NaN`. For example, in Vector2_TEST.rb,
https://github.com/ignitionrobotics/ign-math/blob/080eb34dde41243fded76bbdf67377192f6044ef/src/ruby/Vector2_TEST.rb#L262-L268
`nanVec` is assigned `Vector2d.NaN`, but this assignment only creates a reference to `Vector2d.NaN`, so mutating `nanVec` is equivalent to mutating `Vector2d.NaN` and causes a segfault because the symbol associated with `Vector2d.Nan` is in the read only data section of the shared library.

One solution to change the test to make a copy like so:
```c++
    nanVec = Ignition::Math::Vector2d.new(Ignition::Math::Vector2d.NaN)
```

scpeters:
> One solution to change the test to make a copy like so:
> 
> ```c++
>     nanVec = Ignition::Math::Vector2d.new(Ignition::Math::Vector2d.NaN)
> ```

thanks; I've fixed the ruby tests in https://github.com/ignitionrobotics/ign-math/pull/283/commits/279d973be4c5798715bec8a869385218d4895a54

azeey:
We need a similar change for the failing python tests. Instead of assignment, we need to invoke the copy constructor. e.g.
```python
# Instead of 
nanVec = Vector2d.NAN
# Use
nanVec = Vector2d(Vector2d.NAN)
```

azeey:
I have found a way to make `Vector*.NAN` return a copy instead of a reference such that the tests pass without any modification (after reverting https://github.com/ignitionrobotics/ign-math/commit/279d973be4c5798715bec8a869385218d4895a54). For SWIG, we can use the `%extend` keyword to create a special function for getting the `NaN` constant:

```c++
      public: %extend {
        static Vector2 NaN()
        {
          return ignition::math::Vector2<T>::NaN;
        }
      }
```
The generated code allocates a new `Vector*` object equal to `Vector*::NaN`

For pybind11, we can use the `py::return_value_policy::copy` as described [here](https://pybind11.readthedocs.io/en/stable/advanced/functions.html#return-value-policies)

```c++
    .def_readonly_static("NAN", &Class::NaN, py::return_value_policy::copy,
        "math::Vector2(NaN, NaN)")
```

If that seems reasonable, I can push my changes to this PR.

jwnimmer-tri:
I'm surely not the final word here, but for my part I'd say yes -- for languages that don't enforce `const`, having the global constants avoid mutation accidents by always returning a new copy sounds good to me.

scpeters:
> I'm surely not the final word here, but for my part I'd say yes -- for languages that don't enforce `const`, having the global constants avoid mutation accidents by always returning a new copy sounds good to me.

agreed

scpeters (approved):
looks good to me! I think the complaint about coverage decreasing is a red herring

mjcarroll (approved):
Looks good, thanks for iterating @jwnimmer-tri

PR #284, opened 2021-11-22
Fix windows warnings in Vector2, 3 and 4

ahcorde:
Signed-off-by: ahcorde <ahcorde@gmail.com>
# 🦟 Warnings

## Summary

Fix windows warnings in Vector2, 3 and 4

## Checklist
- x ] Signed all commits for DCO
- [ ] Added tests

### DIFF
--- a/include/ignition/math/Vector2.hh
+++ b/include/ignition/math/Vector2.hh
@@ -108,10 +108,10 @@ namespace ignition
       {
         double d = this->Length();
 
-        if (!equal<T>(d, static_cast<T>(0.0)))
+        if (!equal<T>(static_cast<T>(d), static_cast<T>(0.0)))
         {
-          this->data[0] /= d;
-          this->data[1] /= d;
+          this->data[0] /= static_cast<T>(d);
+          this->data[1] /= static_cast<T>(d);
         }
       }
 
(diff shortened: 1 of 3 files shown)


include/ignition/math/Vector2.hh
```diff
@@ -108,10 +108,10 @@ namespace ignition
       {
         double d = this->Length();
```
jwnimmer-tri, line 109: The `Length()` function returns a `T` already.  It seems to me like _this_ line is where we should apply a fix (by changing the local variable to be of type `d`), not casting it repeatedly in subsequent lines.

include/ignition/math/Vector2.hh
```diff
@@ -128,8 +128,8 @@ namespace ignition
       /// \return the result
       public: Vector2 Round()
       {
-        this->data[0] = nearbyint(this->data[0]);
-        this->data[1] = nearbyint(this->data[1]);
+        this->data[0] = static_cast<T>(nearbyint(this->data[0]));
```
jwnimmer-tri, line 131: The https://en.cppreference.com/w/cpp/numeric/math/nearbyint function returns the same type as its argument, so I'm suspicious of the need for a cast here.  Is the compiler selecting the wrong overload?  Maybe we should be calling `std::nearbyint` instead of using ADL or relying on the C-global namespace lookup?  It also seems like this file is missing the `#include` statement for `<cmath>`, which could also be the problem.

jwnimmer-tri (changes requested):
FYI I'm not a project maintainer, just trying to do my part and review other PRs as requested in the template for the PRs that I have submitted.

jwnimmer-tri (approved):

chapulina (approved):

osrf-triage:

PR #286, opened 2021-11-23
Use nested struct to provide constexpr constants

jwnimmer-tri:
# 🦟 Bug fix

A portion of #269.

## Summary

Use constexpr for simple static constants, to avoid the C++ global static construction and destruction order fiascos.

This is an alternative approach vs #283.  Due to MSVC's apparent limitiations, the only way to provide inline static constants is via a helper struct (or namespace-scoped globals), so that the `class Color` exists as a complete type prior to declaring constants that use it.

Deprecate the const-only (non-constexpr) spelling of the constants.

At the moment, only Color has been refactored with this pattern while we discuss the implications.

In this version of the fix, we have an API+ABI break (with a deprecation transition), and thus will require users to refactor their code.  In the #283 version of the fix, we remain API compatible (but still break ABI).

## Checklist
- [ ] Added tests

### DIFF
--- a/include/ignition/math/Color.hh
+++ b/include/ignition/math/Color.hh
@@ -36,22 +36,25 @@ namespace ignition
     /// (A) component. Each color component is in the range [0..1].
     class IGNITION_MATH_VISIBLE Color
     {
+      // This struct of named constants is defined later in this file.
+      public: struct IGNITION_MATH_VISIBLE Constant;
+
       /// \brief (1, 1, 1)
-      public: static const Color White;
+      public: IGN_DEPRECATED(7) static const Color White;
       /// \brief (0, 0, 0)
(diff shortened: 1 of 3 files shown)


jwnimmer-tri:
I'm seeking input from project maintainers on whether this (#286) or else #283 is your preferred tactic to resolve the fiasco problem. Once I have that, I'll close the dispreferred PR, and update the preferred PR to cover all of the relevant types (Vector2, Vector3, etc.) that use the same pattern.

mjcarroll:
Thanks for taking this on @jwnimmer-tri .  I'm taking a look at both approaches and will let you know what I prefer, at least.

jwnimmer-tri:
Closing in lieu of #283.

PR #294, opened 2021-11-30
Drop unnecessary moving window filter pimpl

mjcarroll:
Came across this as I was deprecating in `ign-common`.  The `pimpl` pattern doesn't make sense for a templated class since the implementation is ultimately exposed to downstream users.

This breaks API/ABI, so I'm targeting main with it.


jwnimmer-tri:
This touches upon a point that I was going to file an issue about later on -- when the math classes here are templated on `<T>`, is that _only_ in support of choosing either `float` or `double` for the scalar type, or is the set of allowed types open to the user to expand?

The documentation isn't clear on what types are permitted to use for `T`.  The code as-written does not seem sufficiently general to allow for user-provided floating point numbers.  (It would need to use either more ADL or Eigen's NumTraits, if more types were wanted.)

Assuming that the library only supports C++'s native floating-point types, then I don't see any reason to place the function definitions in the header files.  The implementation could move into the cc files by using explicit template instantiation over the set of supported template arguments.

mjcarroll:
> is that only in support of choosing either float or double for the scalar type, or is the set of allowed types open to the user to expand?

I think that's a good point.  I wasn't around for the earlier iterations of ign-math, so I'm not sure what the intent was.  I know that there is at least support for `int`, `float`, and `double` for `Vector`, but that I don't believe that `int` would work here with the way it is implemented. 

> Assuming that the library only supports C++'s native floating-point types, then I don't see any reason to place the function definitions in the header files. The implementation could move into the cc files by using explicit template instantiation over the set of supported template arguments.

I think this would be a reasonable approach.

 Perhaps @nkoenig or @chapulina could lend some insight?

chapulina:
I don't have a lot of context on this class, I don't remember ever using it, and I couldn't find any code that uses it. I traced it back to this commit on Gazebo classic: https://github.com/osrf/gazebo/commit/16e1af9aea1ac962a85d6d5c39fcdcfeeea1e6ce

> The implementation could move into the cc files by using explicit template instantiation over the set of supported template arguments.

I think that's ok. There'd be a bit of duplication, but the class is small enough...

mjcarroll:
Added explicit instantiation for:

```
template class MovingWindowFilter<int>;
template class MovingWindowFilter<float>;
template class MovingWindowFilter<double>;
template class MovingWindowFilter<ignition::math::Vector3i>;
template class MovingWindowFilter<ignition::math::Vector3f>;
template class MovingWindowFilter<ignition::math::Vector3d>;
```

mjcarroll:
Blocked on #299

nkoenig:
I don't have additional context on this class. This looks fine.

include/gz/math/MovingWindowFilter.hh
```diff
@@ -19,64 +19,31 @@
...
+    class GZ_MATH_VISIBLE MovingWindowFilter
     {
-      /// \brief Constructor
-      public: MovingWindowFilter();
+      // \brief Constructor
+      public: MovingWindowFilter(unsigned int _windowSize = 4);
 
-      /// \brief Destructor
-      public: virtual ~MovingWindowFilter();
+      // \brief Constructor
```
scpeters, line 45: Destructor

include/gz/math/MovingWindowFilter.hh
```diff
@@ -19,64 +19,31 @@
...
     /// version of MovingWindowFilter in the Gazebo Common library.
     ///
     /// The default window size is 4.
     template< typename T>
-    class MovingWindowFilter
+    class GZ_MATH_VISIBLE MovingWindowFilter
     {
-      /// \brief Constructor
-      public: MovingWindowFilter();
+      // \brief Constructor
```
scpeters, line 42:
```suggestion
      /// \brief Constructor
```

include/gz/math/MovingWindowFilter.hh
scpeters:
```suggestion
      /// \brief Destructor
```
also fix the ///

scpeters (approved):
looks good with clean CI

PR #296, opened 2021-12-02
Add codecheck to focal CI

mjcarroll:
This is a workaround to enable codecheck on ign-math before we turn it on globally for focal.

Since `ign-math7`/`main` is only supported on `focal` forward, it makes sense to turn these checks on.


chapulina:
How about enabling this through the action? If it's centralized, we can flip the switch to make checking the default for all libraries more easily. It also prevents adding a new `between_cmake_make.sh` file to all libraries.

mjcarroll:
I think that makes sense, I had that thought after I opened this PR, moving this back to draft for now.

mjcarroll:
This is now using the branch from ignition-tooling/action-ignition-ci#43.  Once that lands, this will revert back to the `focal` branch of that repo.

mjcarroll:
Tooling branch is merged, this is ready for review

mjcarroll:
Confirmed it worked: 
image

chapulina (approved):

PR #320, opened 2021-12-21
Fix Color::HSV() incorrect hue output

AzulRadio:
# 🦟 Bug fix

## Summary

* **The implementation of Hue of Color::HSV() does not seem right, where HSV stands for hue, saturation, value.**
* OS Version: Ubuntu 20.04
* Source build from b3990351bc8b3b56746db93f996fa746e0759692, branch ign-math6
* Also update the expected behavior in the color tutorial accordingly.
* Change test cases for Color

## Why also change the test cases:

* [Here](https://ignitionrobotics.org/api/math/6.9/classignition_1_1math_1_1Color.html#a00b4e0fa35c908ffd58cf11f25f0ad23) is a document of Ignition Math that said the range of Hue should be [0, 360]. 
* The current Hue implementation does not seem to fit in any HSV standard as far as I know.
* Repeatedly converting back and forth between RGB and HSV will change the value of the color drastically. 

## Description
* `Color::HSV()` in `Color.cc` has an incorrect output hue (index 0 of return `Vector3f` object) 

#### Example 1: Converting the RGB (0.5, 0.6, 0.7) to HSV.
* Expected output: (210, 0.286, 0.7), can verify with [RGB to HSV](https://www.rapidtables.com/convert/color/rgb-to-hsv.html)
* Actual output: (3.5, 0.286, 0.7)

#### Example 2: Converting the RGB (0.0, 0.0, 0.0) to HSV.
* Expected output: (0, 0, 0)
* Actual output: (-1, 0, 0)

#### Examples can be verified by putting the following file in /examples

```
#include <iostream>
#include <ignition/math/Color.hh>

int main() 
{
  ignition::math::Color b = ignition::math::Color(0.0, 0.0, 0.0);
  std::cout << "The RGBA value of b: " << b << std::endl;
  std::cout << "The HSV value of b: " << b.HSV() << std::endl;

  ignition::math::Color a = ignition::math::Color(0.5, 0.6, 0.7);
  std::cout << "The RGBA value of a: " << a << std::endl;
  std::cout << "The HSV value of a: " << a.HSV() << std::endl;
}
```

#### Incorrect Output on terminal
bug

#### Correct Output on terminal
fix

## Checklist
- [ ] Added tests

### DIFF
--- a/Migration.md
+++ b/Migration.md
@@ -11,6 +11,8 @@ release will remove the deprecated code.
    preserve behaviour, the `LOCAL` frame was left with the bug, and a new
    `LOCAL2` frame was introduced, which can be used to get the correct
    calculations.
+   
+2. **Color::HSV()**: A bug related the the hue output of this function was fixed.
 
 ## Ignition Math 4.X to 5.X
 
--- a/Migration.md
+++ b/Migration.md
@@ -5,6 +5,10 @@ Deprecated code produces compile-time warnings. These warning serve as
 notification to users that their code should be upgraded. The next major
 release will remove the deprecated code.
 
+## Ignition Math 6.9.2 to 6.10.0
+
+1. **Color::HSV()**: A bug related to the hue output of this function was fixed.
+
 ## Ignition Math 6.8 to 6.9
 
 1. **SphericalCoordinates**: A bug related to the LOCAL frame was fixed. To
@@ -12,8 +16,6 @@ release will remove the deprecated code.
    `LOCAL2` frame was introduced, which can be used to get the correct
    calculations.
    
-2. **Color::HSV()**: A bug related the the hue output of this function was fixed.
-
 ## Ignition Math 4.X to 5.X
 
 ### Additions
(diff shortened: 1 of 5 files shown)


chapulina (approved):
Thanks for the fix!

In general, we avoid changes in behavior in a stable version even if they're bug fixes, because downstream users may be compensating for it and may be broken by the change. But in this case, I think it's safe to assume that users aren't using this function since it returns wrong values, so fixing it should be good. I think we should add a note to the migration guide just in case.

AzulRadio:
@chapulina Thanks for the feedback! I updated the migration.md but I am not entirely sure how to write this guide (which version are we migrating from & to, etc). Could you help me verify what I added in the guide is correct? Thanks for your time!

chapulina:
> I updated the migration.md but I am not entirely sure how to write this guide (which version are we migrating from & to, etc). Could you help me verify what I added in the guide is correct? 

Thanks, the text looks good to me. But since the fix will go into the next release, mind putting it under a new section `## Ignition Math 6.9.2 to 6.10.0` on the top?

> Thanks for your time!

Thank you too!

osrf-triage:

PR #285, opened 2021-11-22
Modified cmake target name for Ruby interfaces

ahcorde:
Signed-off-by: ahcorde <ahcorde@gmail.com>

# 🦟 Bug fix

this PR blocks this other PR https://github.com/ignitionrobotics/ign-math/pull/280/files

pybind11 requires to call the target with the same name of the exported module. We would like to use `import gazebo.math` with Python interfaces. We are using this target name for the Ruby SWIG modules,  SWIG allows to set some variable to call the module with the desired name.

This is just a change of the target name should not affect behaviour.

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [x] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [x] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge**

PR #327, opened 2021-12-22
Cherry-pick #285, #280 to main

scpeters:
I believe #280 and #285 will help resolve some issues in #283.

Remember to use rebase-and-merge

### DIFF
--- a/.github/ci/packages.apt
+++ b/.github/ci/packages.apt
@@ -4,3 +4,4 @@ python3-distutils
 python3-pybind11
 ruby-dev
 swig
+python3-pybind11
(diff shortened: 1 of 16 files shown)


chapulina:
Cherry-picking will make the commits still show up when we see the branch comparison. Can we do it as a merge forward like previous PRs?

scpeters:
> Cherry-picking will make the commits still show up when we see the branch comparison. Can we do it as a merge forward like previous PRs?

the idea was to help expedite a fix for the test failures in https://github.com/ignitionrobotics/ign-math/pull/283, but if there's not a rush, then I'll just continue merging forward incrementally

azeey (approved):

PR #344, opened 2021-12-27
Merge 6.9.0 to main

scpeters:
# ➡️ Forward port

Port `ign-math6` as of 6.9.0 to `main`. Needed by https://github.com/ignitionrobotics/sdformat/pull/803 and https://github.com/ignition-tooling/release-tools/issues/574

It has compilation warnings that need fixing still

Branch comparison: https://github.com/ignitionrobotics/ign-math/compare/main...ignition-math6_6.9.0

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,6 +69,7 @@ if (SWIG_FOUND)
 
   ########################################
   # Include python
+  find_package(PythonInterp 3 REQUIRED) # change to Python3 when Bionic is EOL
   find_package(PythonLibs QUIET)
   if (NOT PYTHONLIBS_FOUND)
     message (STATUS "Searching for Python - not found.")
--- a/Changelog.md
+++ b/Changelog.md
@@ -46,6 +46,71 @@
 
 ### Ignition Math 6.x.x
 
+## Ignition Math 6.9.0 (2021-09-28)
+
+1. Volume below a plane for spheres and boxes
+    * [Pull request #219](https://github.com/ignitionrobotics/ign-math/pull/219)
+
+1. 🌐 Spherical coordinates: bug fix, docs and sanity checks
+    * [Pull request #235](https://github.com/ignitionrobotics/ign-math/pull/235)
+
+1. Add Vector(2|3|4)<T>::NaN to easily create invalid vectors
+    * [Pull request #222](https://github.com/ignitionrobotics/ign-math/pull/222)
+
+1. Add options to install python/ruby in system standard paths
+    * [Pull request #236](https://github.com/ignitionrobotics/ign-math/pull/236)
+
+1. Add eigen utils to convert mesh 3d vertices to oriented box
+    * [Pull request #224](https://github.com/ignitionrobotics/ign-math/pull/224)
+
+1. Python interface
+
+    1. Adds python interface to RollingMean, Color and Spline
+        * [Pull request #218](https://github.com/ignitionrobotics/ign-math/pull/218)
+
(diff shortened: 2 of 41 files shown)


chapulina (commented):
We need to update the python bindings to accommodate the deprecated functions so the warnings are fixed. Since the python API is still in beta, we don't need to tick-tock it, so I say we just update the SWIG files to use the new functions.

chapulina:
> We need to update the python bindings to accommodate the deprecated functions so the warnings are fixed.

Done in c69f1a9a0594fa9254edfa8c6b142200b0c41d93. One of the tests fails for me locally before and after this PR though. I tried a few different things, and even if I don't use `abs`, the vector's value doesn't look right locally. Let's see what CI says.

```
ERROR: test_math (__main__.TestQuaternion)
----------------------------------------------------------------------
  File "src/ign-math/src/python/Quaternion_TEST.py", line 191, in test_math
    self.assertAlmostEqual(
  File "/usr/lib/python3.8/unittest/case.py", line 943, in assertAlmostEqual
    diff = abs(first - second)
TypeError: bad operand type for abs(): 'Vector3d'

```

scpeters:
> > We need to update the python bindings to accommodate the deprecated functions so the warnings are fixed.
> 
> Done in [c69f1a9](https://github.com/ignitionrobotics/ign-math/commit/c69f1a9a0594fa9254edfa8c6b142200b0c41d93). One of the tests fails for me locally before and after this PR though. I tried a few different things, and even if I don't use `abs`, the vector's value doesn't look right locally. Let's see what CI says.
> 
> ```
> ERROR: test_math (__main__.TestQuaternion)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "src/ign-math/src/python/Quaternion_TEST.py", line 191, in test_math
>     self.assertAlmostEqual(
>   File "/usr/lib/python3.8/unittest/case.py", line 943, in assertAlmostEqual
>     diff = abs(first - second)
> TypeError: bad operand type for abs(): 'Vector3d'
> ```

I fixed a few more instances of deprecated APIs being called in python tests and amended https://github.com/ignitionrobotics/ign-math/commit/c69f1a9a0594fa9254edfa8c6b142200b0c41d93 to https://github.com/ignitionrobotics/ign-math/commit/482381defbdc5224fbc938206a9dfc5098cbbaab and force-pushed

chapulina:
> I fixed a few more instances of deprecated APIs being called in python tests and amended c69f1a9 to 482381d and force-pushed 

It's still failing. I've been looking into it, I think it's because of some of the API changes on `main` for constructors and accessors. For example, this test failure:

```
  FAIL: test_pose (__main__.TestPose3)
  ----------------------------------------------------------------------
    File "/github/workspace/src/python/Pose3_TEST.py", line 44, in test_pose
      self.assertAlmostEqual((B + A).pos().x(), 1.0 + 1.0/math.sqrt(2))
  AssertionError: 0.0 != 1.7071067811865475 within 7 places (1.7071067811865475 difference)
```

Is fixed if we use `(B + A).x()` instead of `(B + A).pos().x()`.

chapulina:
Updated the last commit from 482381d to 84837c2. I changed some tests to workaround the failures, and commented out others. I still don't know what's the issue, but I suspect it's related to the `Vector3` copy constructor. I tried changing that from `default` to an implementation though and that didn't help :woman_shrugging:

chapulina (approved):
Added TODOs in 237fab5 to revisit the quaternion python tests after migrating to pybind11. I think there's little point in improving the SWIG python implementation right now. It may become important when we use it for Ruby though.

PR #366, opened 2022-01-13
Use const instead of constexpr in Ellipsoid constructor

adlarkin:
Signed-off-by: Ashton Larkin <42042756+adlarkin@users.noreply.github.com>

# 🦟 Bug fix

## Summary
While working on https://github.com/ignitionrobotics/sdformat/pull/821, I had the following build error:
```
In file included from /usr/include/ignition/math6/ignition/math/Ellipsoid.hh:132:0,
                 from /sdf_to_usd/src/sdformat/include/sdf/Ellipsoid.hh:20,
                 from /sdf_to_usd/src/sdformat/src/Ellipsoid_TEST.cc:19:
/usr/include/ignition/math6/ignition/math/detail/Ellipsoid.hh: In instantiation of 'T ignition::math::v6::Ellipsoid<Precision>::Volume() const [with Precision = double]':
/sdf_to_usd/src/sdformat/src/Ellipsoid_TEST.cc:27:3:   required from here
/usr/include/ignition/math6/ignition/math/detail/Ellipsoid.hh:99:43: error: '(1.2566370614359172e+1 / 3.0e+0)' is not a constant expression
   constexpr T kFourThirdsPi = 4. * IGN_PI / 3.;
```

I believe the issue is with `IGN_PI`. It is [defined as a floating point value](https://github.com/ignitionrobotics/ign-math/blob/ignition-math6_6.9.3-pre1/include/ignition/math/Helpers.hh#L176-L188), but the c++ standard says the following ([source](https://en.cppreference.com/w/cpp/language/constant_expression#Usable_in_constant_expressions)):
```
a variable is usable in constant expressions at a point P if the variable is: 
- a constexpr variable, or
- it is a constant-initialized variable
    - of reference type or
    - of const-qualified integral or enumeration type
```

So, since `IGN_PI` is not a const-qualified integral/enumeration type, I believe this is causing an error in the sdformat ellipsoid unit test (I'm not sure why this hasn't been an issue until now, though).

## Checklist
- [ ] Added tests

### DIFF
--- a/include/ignition/math/detail/Ellipsoid.hh
+++ b/include/ignition/math/detail/Ellipsoid.hh
@@ -96,7 +96,7 @@ std::optional< MassMatrix3<T> > Ellipsoid<T>::MassMatrix() const
 template<typename T>
 T Ellipsoid<T>::Volume() const
 {
-  constexpr T kFourThirdsPi = 4. * IGN_PI / 3.;
+  const T kFourThirdsPi = 4. * IGN_PI / 3.;
   return kFourThirdsPi * this->radii.X() * this->radii.Y() * this->radii.Z();
 }
 


chapulina (approved):

azeey:
I think it is possible to use `IGN_PI` in a const expression because it's a macro, not a variable. I've tested it on [godbolt](https://godbolt.org/z/TzqjGYcxh) and here's an answer that elaborates how the "const-qualified integral or enumeration type` statement is to be interpreted: https://stackoverflow.com/questions/58514970/constexpr-int-and-constexpr-double-in-c

I suspect there might be something else wrong causing the error. Does the error happen in CI in sdformat?

adlarkin:
> I suspect there might be something else wrong causing the error. Does the error happen in CI in sdformat?

For the sdformat PR I was testing that had this error, CI currently isn't working because the PR adds a new dependency and needs to be updated to use the `ign-cmake` code (https://github.com/ignitionrobotics/sdformat/pull/736). So perhaps it's something in that PR that's causing the issue, but I can't say for sure (the PR is large and I am just now starting to review it). Is there a reason why `constexpr` would be preferred over `const` in this case?

azeey:
The generated code looks exactly the same, so I don't think one would be preferred over the other, but the fact that it's causing an error somewhere else makes me suspect there's something else going on. This code has been released and CI has been running the `Ellipsoid_TEST` in sdformat for a while without any issues, so I'd be curious to know what change caused the error.

adlarkin:
> but the fact that it's causing an error somewhere else makes me suspect there's something else going on. This code has been released and CI has been running the Ellipsoid_TEST in sdformat for a while without any issues, so I'd be curious to know what change caused the error.

I agree. What's weird too is that the PR I was testing does not touch the ellipsoid code or test at all. The ultimate goal is to break up https://github.com/ignitionrobotics/sdformat/pull/736 into several smaller PRs (like what's being done with https://github.com/ignitionrobotics/sdformat/pull/817), so once that happens, maybe we can catch anything strange that's going on in these new sdformat additions.

osrf-triage:

PR #372, opened 2022-01-28
[math7] Tidy up unused variables

mjcarroll:
Uncovered a few things during a clang-12 build.

The big one is that we weren't correctly forwarding arguments in `Quaternion::Axis`, so I fixed it and added a test to prevent further regressions.

Everything else was unused variables.

### DIFF
--- a/include/ignition/math/Quaternion.hh
+++ b/include/ignition/math/Quaternion.hh
@@ -303,7 +303,7 @@ namespace ignition
       /// \deprecated Use SetFromAxisAngle(T, T, T, T)
       public: void IGN_DEPRECATED(7) Axis(T _ax, T _ay, T _az, T _aa)
       {
-        this->SetFromAxisAngle(_ax, _ay, _ax, _aa);
+        this->SetFromAxisAngle(_ax, _ay, _az, _aa);
       }
 
       /// \brief Set the quaternion from an axis and angle.
@@ -716,7 +716,6 @@ namespace ignition
       /// \param[in] _scale Amount to scale this quaternion
       public: void Scale(T _scale)
       {
-        Quaternion<T> b;
         Vector3<T> axis;
         T angle;
 
--- a/src/PID_TEST.cc
+++ b/src/PID_TEST.cc
@@ -108,40 +108,6 @@ TEST(PidTest, SetValues)
 }
 
 /////////////////////////////////////////////////
-TEST(PidTest, EqualOperatorCornerCase)
-{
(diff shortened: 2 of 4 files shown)


src/Quaternion_TEST.cc
```diff
@@ -324,6 +324,21 @@ TEST(QuaternionTest, MathAxis)
 {
   math::Quaterniond q(IGN_PI*0.1, IGN_PI*0.5, IGN_PI);
 
+#ifndef _WIN32
```
azeey, line 327: Recommend `IGN_UTILS_WARN_IGNORE__DEPRECATED_DECLARATION`.

src/PID_TEST.cc
```diff
@@ -121,7 +121,14 @@ TEST(PidTest, EqualOperatorCornerCase)
   EXPECT_DOUBLE_EQ(pid.CmdOffset(), 1.23);
   EXPECT_DOUBLE_EQ(pid.Cmd(), 0.0);
 
+#ifndef _WIN32
+# pragma GCC diagnostic push
+# pragma GCC diagnostic ignored "-Wself-assign-overloaded"
```
azeey, line 126: I'm getting a warning about this: `warning: unknown option after ‘#pragma GCC diagnostic’ kind [-Wpragmas]` on gcc 9.3. Maybe it's only recognized in clang?

include/ignition/math/Quaternion.hh
```diff
@@ -303,7 +303,7 @@ namespace ignition
       /// \deprecated Use SetFromAxisAngle(T, T, T, T)
       public: void IGN_DEPRECATED(7) Axis(T _ax, T _ay, T _az, T _aa)
       {
-        this->SetFromAxisAngle(_ax, _ay, _ax, _aa);
+        this->SetFromAxisAngle(_ax, _ay, _az, _aa);
```
azeey, line 306: I get a warning about this (before this PR) with gcc 9. I looked in our CI and even though its running on Focal now, we still install and use gcc8.

src/PID_TEST.cc
mjcarroll: Looking at it again, I believe this whole test is irrelevant since we switch to the `ImplPtr` pattern.  I have removed it.

include/ignition/math/Quaternion.hh
mjcarroll: I think it's probably time to consider using the default on focal, rather than using the older gcc version.  We only needed gcc8 over gcc7.

azeey: I agree. Default makes sense on focal.

azeey (approved):

PR #315, opened 2021-12-17
Add GaussMarkovProcess pybind11 interface and examples

ahcorde:
Signed-off-by: ahcorde <ahcorde@gmail.com>

# 🎉 New feature

## Summary

This PR creates a GaussMarkovProcess Pybind11 interface and adds examples

## Checklist
- [x] Signed all commits for DCO
- [x] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [x] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [x] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #361, opened 2022-01-05
Fix pybind11 issues with macOS

scpeters:
This fixes a test failure and compiler error when running the pybind11 tests on macOS. This pull request has a matching branch name with https://github.com/osrf/homebrew-simulation/pull/1793, so the CI is using `pybind11` and is now clean.

PR #379, opened 2022-02-25
Merge 6 -> 7 as of pybind11 GaussMarkovProcess

scpeters:
# ➡️ Forward port

Port ign-math6 to main as of #315.

Branch comparison: https://github.com/ignitionrobotics/ign-math/compare/main...7717f4913fb407f8e4a69203aace0a2b47799956


ahcorde (changes requested):
we have a warning related with `Quaternion<int>` I don't know if does make sense or not to have a class like this. Few can only represents with this class a few angles.

scpeters:
> we have a warning related with `Quaternion<int>` I don't know if does make sense or not to have a class like this. Few can only represents with this class a few angles.

I agree it doesn't make much sense, but it might be needed for `Matrix3i` due to the coupling between Matrix3 and Quaternion

I've cherry-picked the fix for the warning from part of #361

ahcorde (approved):

PR #387, opened 2022-03-15
Support polynomial piecewise scalar fields

hidmic:
# 🎉 New feature

## Summary
Spin-off from https://github.com/ignitionrobotics/ign-gazebo/pull/1357. This PR introduces a number of classes, namely:

- an Interval class for scalar intervals
- a Polynomial3 class for cubic polynomials
- a Region3 class for regions in R^3
- a SeparableScalarField3 class for separable scalar fields in R^3
- a PiecewiseScalarField3 class for piecewise scalar fields in R^3

to support polynomial piecewise scalar fields.

## Checklist
- [x] Signed all commits for DCO
- [x] Added tests
- [x] Added example and/or tutorial
- [x] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [x] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #388, opened 2022-03-16
Add Interval class

hidmic:
# 🎉 New feature

## Summary
Split from #387. This PR introduces an `Interval` class to represent scalar, real intervals.

## Checklist
- [x] Added tests

### DIFF
--- a/include/ignition/math/Interval.hh
+++ b/include/ignition/math/Interval.hh
@@ -1,0 +1,292 @@
+/*
+ * Copyright (C) 2022 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+#ifndef IGNITION_MATH_INTERVAL_HH_
+#define IGNITION_MATH_INTERVAL_HH_
+
+#include <cmath>
+#include <limits>
+#include <ostream>
+#include <type_traits>
+#include <utility>
+
+#include <ignition/math/config.hh>
+
+namespace ignition
+{
+  namespace math
+  {
+    // Inline bracket to help doxygen filtering.
+    inline namespace IGNITION_MATH_VERSION_NAMESPACE {
+    //
+    /// \class Interval Interval.hh ignition/math/Interval.hh
+    /// \brief The Interval class represents a range of real numbers.
+    /// Intervals may be open (a, b), left-closed [a, b), right-closed
+    /// (a, b], or fully closed [a, b].
+    ///
+    /// ## Example
+    ///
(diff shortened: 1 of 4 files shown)


hidmic:
@chapulina PTAL!

ahcorde (changes requested):
Do you mind to add this a Python interfaces ? or if you don't have time please open an issue.

hidmic:
> Do you mind to add this a Python interfaces ? or if you don't have time please open an issue.

@ahcorde I'll open an issue once we merge this PR. I promise I'll get back and implement `pybind11` bindings later on, but I need this feature in ASAP.

include/ignition/math/Interval.hh
```diff
@@ -0,0 +1,292 @@
...
+        }
+        return true;
+      }
+
+      /// \brief Equality test operator
+      /// \param _other interval to check for equality
+      /// \return true if intervals are equal, false otherwise
+      public: bool operator==(const Interval<T> &_other) const
+      {
+        return this->Contains(_other) && _other.Contains(*this);
```
chapulina, line 244: Just curious, this results in more comparisons than directly comparing the 4 member variables. Does this cover a case that wouldn't be covered otherwise, or is it just a more formal way of comparing intervals?

chapulina (approved):
This looks great, thanks for the tests, documentation and example. I just have minor comments, but nothing blocking.

include/ignition/math/Interval.hh
```diff
@@ -0,0 +1,292 @@
...
+
+namespace ignition
+{
+  namespace math
+  {
+    // Inline bracket to help doxygen filtering.
+    inline namespace IGNITION_MATH_VERSION_NAMESPACE {
+    //
+    /// \class Interval Interval.hh ignition/math/Interval.hh
+    /// \brief The Interval class represents a range of real numbers.
```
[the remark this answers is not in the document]
hidmic: @chapulina first and foremost, I had no need for them and there's no easy way to represent unbounded integer intervals. Integer intervals also need special treatment e.g. (0, 10) and [1, 9] are equivalent when defined over integers but if defined over the reals. 

I can also imagine someone attempting to use modular arithmetic integer intervals, so I (naively) think it's best for Ignition Math to deal with integer intervals separately.

include/ignition/math/Interval.hh
hidmic: @chapulina this allows me to define equality without specifying an arbitrary tolerance. Checking floating-point numbers for equality has a few rough edges (and any attempt to do so generates a warning when building Ignition Math).

include/ignition/math/Interval.hh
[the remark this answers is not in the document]
chapulina: Got it, thanks for explaining the caveats

include/ignition/math/Interval.hh
chapulina: Makes sense, thanks!

include/ignition/math/Interval.hh
```diff
@@ -0,0 +1,292 @@
...
+      private: T rightValue{0};
+      /// \brief Whether the interval is left-closed or not
+      private: bool leftClosed{false};
+      /// \brief Whether the interval is right-closed or not
+      private: bool rightClosed{false};
+    };
+
+    namespace detail {
+       template<typename T>
+       const Interval<T> gUnboundedInterval =
```
scpeters, line 279: there may be some issues with static initialization of this variable (see #269). It looks like this is almost following the pattern from #283, but it is not a `constexpr`, so I think some additional work will be needed

PR #390, opened 2022-03-18
Add Region3 class

hidmic:
# 🎉 New feature

## Summary
Split from https://github.com/ignitionrobotics/ign-math/pull/387. This PR introduces a Region3 class to represent regions in R^3 space. To do this, it builds on top of the Interval class introduced by https://github.com/ignitionrobotics/ign-math/pull/388.

## Checklist
- [x] Added tests

### DIFF
--- a/include/ignition/math/Region3.hh
+++ b/include/ignition/math/Region3.hh
@@ -18,6 +18,7 @@
 #define IGNITION_MATH_REGION3_HH_
 
 #include <cmath>
+#include <limits>
 #include <ostream>
 #include <utility>
 
@@ -39,11 +40,15 @@ namespace ignition
     /// an intersection of halfspaces. Regions may be open or
     /// closed in their boundaries, if any.
     ///
+    /// Note that the Region3 class is essentially a set R ⊆ R^3.
+    /// For 3D solid box semantics, use the `AxisAlignedBox` class
+    /// instead.
+    ///
     /// ## Example
     ///
     /// \snippet examples/region3_example.cc complete
     template <typename T>
-    class IGNITION_MATH_VISIBLE Region3
+    class Region3
     {
       /// \brief An unbounded region (-∞, ∞) ✕ (-∞, ∞) ✕ (-∞, ∞)
       public: static const Region3<T> &Unbounded;
@@ -184,9 +189,12 @@ namespace ignition
     namespace detail {
        template<typename T>
        const Region3<T> gUnboundedRegion3(
-           Interval<T>::Unbounded,
-           Interval<T>::Unbounded,
-           Interval<T>::Unbounded);
+           Interval<T>::Open(-std::numeric_limits<T>::infinity(),
+                             std::numeric_limits<T>::infinity()),
+           Interval<T>::Open(-std::numeric_limits<T>::infinity(),
+                             std::numeric_limits<T>::infinity()),
+           Interval<T>::Open(-std::numeric_limits<T>::infinity(),
+                             std::numeric_limits<T>::infinity()));
     }
     template<typename T>
     const Region3<T> &Region3<T>::Unbounded = detail::gUnboundedRegion3<T>;
--- a/include/ignition/math/Region3.hh
+++ b/include/ignition/math/Region3.hh
@@ -1,0 +1,200 @@
+/*
+ * Copyright (C) 2022 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


hidmic:
@chapulina PTAL!

include/ignition/math/Region3.hh
```diff
@@ -0,0 +1,200 @@
...
+
+    namespace detail {
+       template<typename T>
+       const Region3<T> gUnboundedRegion3(
+           Interval<T>::Unbounded,
+           Interval<T>::Unbounded,
+           Interval<T>::Unbounded);
+    }
+    template<typename T>
+    const Region3<T> &Region3<T>::Unbounded = detail::gUnboundedRegion3<T>;
```
chapulina, line 192: Windows doesn't like this:

```
C:\Jenkins\workspace\ign_math-pr-win\ws\ign-math\include\ignition/math/Region3.hh(192): message : while compiling class template static data member 'const ignition::math::v7::Region3<double> &ignition::math::v7::Region3<double>::Unbounded' [C:\Jenkins\workspace\ign_math-pr-win\ws\build\ignition-math7\src\UNIT_Region3_TEST.vcxproj]

C:\Jenkins\workspace\ign_math-pr-win\ws\ign-math\src\Region3_TEST.cc(27): message : see reference to class template instantiation 'ignition::math::v7::Region3<double>' being compiled [C:\Jenkins\workspace\ign_math-pr-win\ws\build\ignition-math7\src\UNIT_Region3_TEST.vcxproj]

C:\Jenkins\workspace\ign_math-pr-win\ws\ign-math\include\ignition/math/Region3.hh(192,47): error C2491: 'ignition::math::v7::Region3<double>::Unbounded': definition of dllimport static data member not allowed [C:\Jenkins\workspace\ign_math-pr-win\ws\build\ignition-math7\src\UNIT_Region3_TEST.vcxproj]
```

include/ignition/math/Region3.hh
```diff
@@ -0,0 +1,200 @@
...
+  {
+    // Inline bracket to help doxygen filtering.
+    inline namespace IGNITION_MATH_VERSION_NAMESPACE {
+    //
+    /// \class Region3 Region3.hh ignition/math/Region3.hh
+    /// \brief The Region3 class represents the cartesian product
+    /// of intervals Ix ✕ Iy ✕ Iz, one per axis, yielding an
+    /// axis-aligned region of R^3 space. It can be thought of as
+    /// an intersection of halfspaces. Regions may be open or
+    /// closed in their boundaries, if any.
```
chapulina, line 40: I can imagine users being confused as to whether they should use this class or `AxisAlignedBox`. Do you think it's worth explaining why one would choose one over the other?

include/ignition/math/Region3.hh
```diff
@@ -0,0 +1,200 @@
...
+      }
+
+      /// \brief Stream insertion operator
+      /// \param _out output stream
+      /// \param _r Region3 to output
+      /// \return the stream
+      public: friend std::ostream &operator<<(
+        std::ostream &_out, const ignition::math::Region3<T> &_r)
+      {
+        return _out <<_r.ix << " x " << _r.iy << " x " << _r.iz;
```
chapulina, line 173: I haven't found any references for the notation of multi-dimensional intervals. Is this some well-known notation? I'm wondering whether instead of `I x I x I`, we should use something like `I I I`, `I, I, I`, `I ✕ I ✕ I` or something else.

chapulina (changes requested):
The general idea looks good to me. I have minor comments about documentation and notation, and the only blocker at the moment is CI.

include/ignition/math/Region3.hh
hidmic: > Is this some well-known notation?

@chapulina I wouldn't call it well-known, no. Having said that, I can't think of a simpler notation than a cartesian product of coordinate intervals. I used `x` instead of `✕` to keep unicode out of terminals, but perhaps that's a non-issue.

include/ignition/math/Region3.hh
hidmic: @chapulina https://github.com/ignitionrobotics/ign-math/pull/390/commits/b36e834a9b16521922f6ac6c0d936a9588fba8d1 should get rid of this.

include/ignition/math/Region3.hh
hidmic: Done in d7c74a5d0b181932eae68f6313184636ac0e8b52.

include/ignition/math/Region3.hh
chapulina: Thanks for the clarification. I'm ok with this since I don't think there's another standard we can use instead.

chapulina (approved):

PR #397, opened 2022-03-25
Add AdditivelySeparableScalarField3 class

hidmic:
# 🎉 New feature

## Summary
Split from https://github.com/ignitionrobotics/ign-math/pull/387. This PR introduces an `AdditivelySeparableScalarField3` class, a particularly simple kind of scalar field in R^3.

## Checklist
- [x] Signed all commits for DCO
- [x] Added tests
- [x] Added example and/or tutorial
- [x] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [x] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #398, opened 2022-03-25
Add PiecewiseScalarField3 class

hidmic:
# 🎉 New feature

## Summary
Split from https://github.com/ignitionrobotics/ign-math/pull/387. This PR introduces a `PiecewiseScalarField3` class to represent a scalar field in R^3 defined on a per region basis.

Depends on #397 for testing.

## Checklist
- [x] Added tests

### DIFF
--- a/include/ignition/math/PiecewiseScalarField3.hh
+++ b/include/ignition/math/PiecewiseScalarField3.hh
@@ -57,7 +57,7 @@ namespace ignition
     ///
     /// \snippet examples/piecewise_scalar_field3_example.cc complete
     template<typename ScalarField3T, typename ScalarT>
-    class IGNITION_MATH_VISIBLE PiecewiseScalarField3
+    class PiecewiseScalarField3
     {
       /// \brief A scalar field P in R^3 and
       /// the region R in which it is defined
--- a/include/ignition/math/PiecewiseScalarField3.hh
+++ b/include/ignition/math/PiecewiseScalarField3.hh
@@ -71,9 +71,9 @@ namespace ignition
 
       /// \brief Constructor
       /// \param[in] _pieces scalar fields Pn and the regions Rn
-      ///   in which these are defined. Regions may not overlap.
-      public: explicit PiecewiseScalarField3(std::vector<Piece> _pieces)
-      : pieces(std::move(_pieces))
+      ///   in which these are defined. Regions should not overlap.
+      public: explicit PiecewiseScalarField3(const std::vector<Piece> &_pieces)
+      : pieces(_pieces)
       {
         for (size_t i = 0; i < pieces.size(); ++i)
         {
@@ -100,6 +100,8 @@ namespace ignition
       }
 
       /// \brief Define piecewise scalar field as `_field` throughout R^3 space
+      /// \param[in] _field a scalar field in R^3
+      /// \return `_field` as a piecewise scalar field
       public: static PiecewiseScalarField3 Throughout(ScalarField3T _field)
       {
         return PiecewiseScalarField3<ScalarField3T, ScalarT>({
@@ -114,7 +116,8 @@ namespace ignition
       {
         auto it = std::find_if(
             this->pieces.begin(), this->pieces.end(),
-            [&](const Piece &piece) {
+            [&](const Piece &piece)
+            {
               return piece.region.Contains(_p);
             });
         if (it == this->pieces.end())
@@ -146,7 +149,8 @@ namespace ignition
       ///   constructed)
       public: ScalarT Minimum(Vector3<ScalarT> &_pMin) const
       {
-        if (this->pieces.empty()) {
+        if (this->pieces.empty())
+        {
           _pMin = Vector3<ScalarT>::NaN;
           return std::numeric_limits<ScalarT>::quiet_NaN();
         }
--- a/include/ignition/math/PiecewiseScalarField3.hh
+++ b/include/ignition/math/PiecewiseScalarField3.hh
@@ -1,0 +1,200 @@
+/*
+ * Copyright (C) 2022 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


include/ignition/math/PiecewiseScalarField3.hh
```diff
@@ -0,0 +1,215 @@
...
+        Region3<ScalarT> region;
+        ScalarField3T field;
+      };
+
+      /// \brief Constructor
+      public: PiecewiseScalarField3() = default;
+
+      /// \brief Constructor
+      /// \param[in] _pieces scalar fields Pn and the regions Rn
+      ///   in which these are defined. Regions may not overlap.
```
chapulina, line 74: When I first read this I understood that regions may or may not overlap. But seeing the error message below, I guess this should be:

```suggestion
      ///   in which these are defined. Regions should not overlap.
```

?

Alternatively, it looks like there's a resolution if the regions do overlap. So we could remove the "Detected overlap..." error message and document that

```suggestion
      ///   in which these are defined. When regions overlap, the first one takes precedence.
```

What do you think?

include/ignition/math/PiecewiseScalarField3.hh
```diff
@@ -0,0 +1,215 @@
...
+                        << pieces[j].region << "). Region #" << i
+                        << " will take precedence when overlapping."
+                        << std::endl;
+            }
+          }
+        }
+      }
+
+      /// \brief Define piecewise scalar field as `_field` throughout R^3 space
+      public: static PiecewiseScalarField3 Throughout(ScalarField3T _field)
```
chapulina, line 103: Document `param[in]` and `return`

And we should make sure we address #269 here too

include/ignition/math/PiecewiseScalarField3.hh
```diff
@@ -0,0 +1,215 @@
...
+
+      /// \brief Evaluate the piecewise scalar field at `_p`
+      /// \param[in] _p piecewise scalar field argument
+      /// \return the result of evaluating `F(_p)`, or NaN
+      ///   if the scalar field is not defined at `_p`
+      public: ScalarT Evaluate(const Vector3<ScalarT> &_p) const
+      {
+        auto it = std::find_if(
+            this->pieces.begin(), this->pieces.end(),
+            [&](const Piece &piece) {
```
chapulina, line 117:
```suggestion
            [&](const Piece &piece)
            {
```

include/ignition/math/PiecewiseScalarField3.hh
```diff
@@ -0,0 +1,215 @@
...
+      /// continuity in the boundaries between regions, if any.
+      /// \param[out] _pMin scalar field argument that yields
+      ///   the minimum, or NaN if the scalar field is not
+      ///   defined anywhere (i.e. default constructed)
+      /// \return the scalar field minimum, or NaN if the
+      ///   scalar field is not defined anywhere (i.e. default
+      ///   constructed)
+      public: ScalarT Minimum(Vector3<ScalarT> &_pMin) const
+      {
+        if (this->pieces.empty()) {
```
chapulina, line 149:
```suggestion
        if (this->pieces.empty())
        {
```

chapulina (commented):
LGTM, just some minor comments

include/ignition/math/PiecewiseScalarField3.hh
hidmic: > Alternatively, it looks like there's a resolution if the regions do overlap.

@chapulina I would've thrown an exception, but it seems you guys have done an effort in keeping `ignition-math` mostly exception-safe.

The issue with establishing a precedence is that it creates a discrepancy between definition and behavior. You'll inspect or print the scalar field and you'll see cases for when, say, `x < 1` and `x > 1` that behave as if you had set them to `x < 1` and `x >= 1`.

include/ignition/math/PiecewiseScalarField3.hh
hidmic: > And we should make sure we address https://github.com/ignitionrobotics/ign-math/issues/269 here too

Hmm, why? It's a function, not a variable.

include/ignition/math/PiecewiseScalarField3.hh
```diff
@@ -0,0 +1,215 @@
...
+        ScalarField3T field;
+      };
+
+      /// \brief Constructor
+      public: PiecewiseScalarField3() = default;
+
+      /// \brief Constructor
+      /// \param[in] _pieces scalar fields Pn and the regions Rn
+      ///   in which these are defined. Regions may not overlap.
+      public: explicit PiecewiseScalarField3(std::vector<Piece> _pieces)
```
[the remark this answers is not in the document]
hidmic: @chapulina it's one way to avoid having to overload the constructor for both const T& and T&&. We can do const T& too, it's cheap anyways.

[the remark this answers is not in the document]
hidmic: Switch to const T& in 38cbc349121a83d2b94df6328f581fa38aeb127e.

include/ignition/math/PiecewiseScalarField3.hh
hidmic: Rephrased documentation in 38cbc349121a83d2b94df6328f581fa38aeb127e

include/ignition/math/PiecewiseScalarField3.hh
hidmic: > Document `param[in]` and `return`

Done in 38cbc349121a83d2b94df6328f581fa38aeb127e

include/ignition/math/PiecewiseScalarField3.hh
hidmic: Done in 38cbc349121a83d2b94df6328f581fa38aeb127e.

I wonder why `codecheck` didn't catch it. BTW is there a styleguide somewhere?

include/ignition/math/PiecewiseScalarField3.hh
hidmic: Done in 38cbc349121a83d2b94df6328f581fa38aeb127e.

include/ignition/math/PiecewiseScalarField3.hh
chapulina: > Hmm, why? It's a function, not a variable.

Ah because I wasn't paying attention 🤦🏽‍♀️

include/ignition/math/PiecewiseScalarField3.hh
chapulina: Yeah it's here:

https://ignitionrobotics.org/docs/all/contributing#style-guides

This kind of thing isn't caught by our linters yet, there's some discussion about this here: https://github.com/ignitionrobotics/ign-gazebo/issues/914

chapulina (approved):
Thanks for addressing the comments!

PR #406, opened 2022-04-08
Adds a volumetric grid field to help with lookup in 3D.

arjo129:

# 🎉 New feature

Closes #<NUMBER>

## Summary
The following PR adds a datastructure that represents a sparse volumetric grid. It allows for storing data in 3D.

# TODO
- [ ] Write unit tests
- [ ] docs

## Checklist
- [ ] Added tests

### DIFF
--- a/include/ignition/math/VolumetricGridLookupField.hh
+++ b/include/ignition/math/VolumetricGridLookupField.hh
@@ -19,6 +19,7 @@
 #define IGNITION_MATH_VOLUMETRIC_GRID_LOOKUP_FIELD_HH_
 
 #include <vector>
+#include <optional>
 
 #include <ignition/math/Vector3.hh>
 
--- a/include/ignition/math/VolumetricGridLookupField.hh
+++ b/include/ignition/math/VolumetricGridLookupField.hh
@@ -1,0 +1,128 @@
+/*
+ * Copyright (C) 2022 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#ifndef IGNITION_MATH_VOLUMETRIC_GRID_LOOKUP_FIELD_HH_
+#define IGNITION_MATH_VOLUMETRIC_GRID_LOOKUP_FIELD_HH_
+
+#include <vector>
+
+#include <ignition/math/Vector3.hh>
+
+#include <ignition/math/detail/AxisIndex.hh>
+
+namespace ignition
+{
+  namespace math
+  {
+    // Inline bracket to help doxygen filtering.
+    inline namespace IGNITION_MATH_VERSION_NAMESPACE {
+      template<typename T>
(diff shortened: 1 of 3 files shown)


include/ignition/math/VolumetricGridLookupField.hh
```diff
@@ -0,0 +1,128 @@
...
+namespace ignition
+{
+  namespace math
+  {
+    // Inline bracket to help doxygen filtering.
+    inline namespace IGNITION_MATH_VERSION_NAMESPACE {
+      template<typename T>
+      /// \brief Lookup table for a volumetric dataset. This class is used to
+      /// lookup indices for a large dataset thats organized in a grid. This
+      /// class is not meant to be used by non-grid like data sets. The grid
```
hidmic, line 37: @arjo129 nit: 
```suggestion
      /// class is not meant to be used with non-grid like datasets. The grid
```

include/ignition/math/VolumetricGridLookupField.hh
```diff
@@ -0,0 +1,128 @@
...
+            index_table[i].resize(num_y);
+            for(int j = 0; j < num_y; ++j)
+            {
+              index_table[i][j].resize(num_x);
+            }
+          }
+
+          for(std::size_t i = 0; i < _cloud.size(); ++i)
+          {
+            auto pt = _cloud[i];
```
hidmic, line 83: @arjo129 nit: use a const reference here

include/ignition/math/VolumetricGridLookupField.hh
```diff
@@ -0,0 +1,128 @@
...
+            std::size_t x_index = x_indices_by_lat.GetIndex(pt.X()).value();
+            std::size_t y_index = y_indices_by_lon.GetIndex(pt.Y()).value();
+            std::size_t z_index = z_indices_by_depth.GetIndex(pt.Z()).value();
```
hidmic, line 86: @arjo129 nit: all can be made `const`

include/ignition/math/detail/AxisIndex.hh
```diff
@@ -0,0 +1,121 @@
...
+          // equal to the value.
+          auto it = axisIndex.lower_bound(_value);
+          if (it == axisIndex.end())
+          {
+            // Out of range
+            return {};
+          }
+          else if (fabs(it->first - _value) < _tol)
+          {
+            // Exact match
```
hidmic, line 101: @arjo129 meta: why introduce a tolerance to check for almost exact matches when we can return all points and defer to the interpolation algorithm to judge?

include/ignition/math/VolumetricGridLookupField.hh
```diff
@@ -0,0 +1,128 @@
...
+            index_table[z_index][y_index][x_index] = i;
```
hidmic, line 87: @arjo129 meta: should we be checking for duplicates?

hidmic (commented):
A humble first pass until @scpeters gets to this.

include/ignition/math/detail/AxisIndex.hh
arjo129: So if you see how the `GetInterpolators` works it returns a variable number of points. By doing this this way it makes for cleaner interpolation code as we can separate the kind of interpolation we want to use (none, linear, bilinear, trilinear) just based on the length of the return vector. If we were to go with the other alternative we would spend a lot more time debugging the corner cases of the interpolation.

hidmic: @arjo129 hmm, why special casing? Isn't linear (resp. bilinear) interpolation just trilinear interpolation for a point that is on an edge (resp. face) of the 3D volumes that make up its grid? I guess I'm missing some numerical nuances here. Anywho, so long as the tolerance can be adjusted I'm OK with this.

arjo129: See #412 for my original plan.

PR #419, opened 2022-05-03
ign -> gz Header Migration : ign-math (2)

methylDragon:
Follow up to https://github.com/ignitionrobotics/ign-math/pull/413
See ignition-tooling/release-tools#711


scpeters (approved):
thanks for the quick fix; looks good, and you can optionally amend the last commit with `(#419)` in the commit message

chapulina (approved):

PR #422, opened 2022-05-05
Fix return policies for some member functions

azeey:
# 🦟 Bug fix

## Summary

We encountered some issues with the python bindings while working on https://github.com/ignitionrobotics/gz-mujoco/pull/5 where calling `link.inertia().mass_matrix()` results in a newly constructed `MassMatrix3d` object instead of returning the object inside the link's `Inertiald` object.

To fix this, generally, member functions that return a const ref should either use the default return policy (which makes a copy) or `py::return_value_policy::reference_internal`. Since we're dealing with small objects, making a copy seems the simplest and safest approach.

For member functions that return a mutable reference, we should use almost always use `py::return_value_policy::reference_internal`. In such instances, if there are const and non-const overloads, the non-const overload should be picked for binding to python.

## Checklist
- [ ] Added tests


scpeters:
are there tests we are missing that would confirm the intended behavior? we don't need to add it now, but I'm wondering what those tests would look like or if some of the classes already have them

azeey:
> are there tests we are missing that would confirm the intended behavior? we don't need to add it now, but I'm wondering what those tests would look like or if some of the classes already have them

I don't know how to test this behavior. It has to do with object lifetime and the python garbage collector, so I'm not sure if there's an easy way to test it.

scpeters:
> > are there tests we are missing that would confirm the intended behavior? we don't need to add it now, but I'm wondering what those tests would look like or if some of the classes already have them
> 
> I don't know how to test this behavior. It has to do with object lifetime and the python garbage collector, so I'm not sure if there's an easy way to test it.

ok, I don't fully understand the issue so I won't approve this, but I think it's fine to merge since it has one approval

azeey:
> > > are there tests we are missing that would confirm the intended behavior? we don't need to add it now, but I'm wondering what those tests would look like or if some of the classes already have them
> > 
> > 
> > I don't know how to test this behavior. It has to do with object lifetime and the python garbage collector, so I'm not sure if there's an easy way to test it.
> 
> ok, I don't fully understand the issue so I won't approve this, but I think it's fine to merge since it has one approval

I can demonstrate the issue from sdformat
```python
import sdformat as sdf
from ignition.math import Inertiald, MassMatrix3d, Vector3d, Pose3d

link = sdf.Link()
mass = 10.0
inertial = Inertiald(MassMatrix3d(mass, Vector3d(2, 3, 4), Vector3d.ZERO),
                     Pose3d())
link.set_inertial(inertial)
print("mass from inertial:", inertial.mass_matrix().mass())
print("mass from link:", link.inertial().mass_matrix().mass())
```

will print 
```
mass from inertial: 10.0
mass from link: 0.0
```
but we expect both to be 10. This PR fixes that.

scpeters:
ok, I think I see what's happening. I still think it would be best to have tests that confirm what we think the API should be doing. Here is a branch where I've added some test expectations to `Inertial_TEST.py` to study the behavior change introduced by this PR:

https://github.com/gazebosim/gz-math/compare/scpeters/python_inertial_tests?expand=1

* https://github.com/gazebosim/gz-math/commit/29cd3fea763572d1e5ee2abcfa1fe6f4b9210d1a: behavior before this PR, calling `inertial.mass_matrix().set_mass()` would change the mass in the object stored by the `inertial` since `mass_matrix()` was returning a reference (tests pass)
* https://github.com/gazebosim/gz-math/commit/997b2c3de34c3d47b9e4424bec4ff2a354823554: merge with `ign-math6` including the changes from this PR (test fails since `mass_matrix()` no longer returns a reference
* https://github.com/gazebosim/gz-math/commit/d793089bc81681b2f5496930582591820e8790d1: change test expectation so it doesn't expect parent object to be changed (tests pass)

this more closely matches the c++ API; hopefully it is pythonic enough

scpeters:
ok, I think I see what's happening. I still think it would be best to have tests that confirm what we think the API should be doing. Here is a branch where I've added some test expectations to `Inertial_TEST.py` to study the behavior change introduced by this PR:

https://github.com/gazebosim/gz-math/compare/scpeters/python_inertial_tests?expand=1

* https://github.com/gazebosim/gz-math/commit/29cd3fea763572d1e5ee2abcfa1fe6f4b9210d1a: behavior before this PR, calling `inertial.mass_matrix().set_mass()` would change the mass in the object stored by the `inertial` since `mass_matrix()` was returning a reference (tests pass)
* https://github.com/gazebosim/gz-math/commit/997b2c3de34c3d47b9e4424bec4ff2a354823554: merge with `ign-math6` including the changes from this PR (test fails since `mass_matrix()` no longer returns a reference
* https://github.com/gazebosim/gz-math/commit/d793089bc81681b2f5496930582591820e8790d1: change test expectation so it doesn't expect parent object to be changed (tests pass)

this more closely matches the c++ API; hopefully it is pythonic enough

azeey:
The C++ member function `Inertial::MassMatrix` returns a const ref so mutation through the ref is not the intended API. As stated in the [pybind11](https://pybind11.readthedocs.io/en/stable/limitations.html) documentation, pybind11 casts away constness in return values, so you don't get an error when doing `inertial.mass_matrix().set_mass()`, but I'm not sure if that's a well defined behavior. And since the intended API in C++ is to use `Inerital::SetMassMatrix` for mutating the mass matrix, I would propose we match that API in python.

PR #424, opened 2022-05-06
MassMatrix3: fix bug in PrincipalAxesOffset tolerances

scpeters:
# 🦟 Bug fix

Fixes a bug in the tolerances used by `MassMatrix3::PrincipalAxesOffset`

## Summary

The `MassMatrix3::PrincipalAxesOffset` API accepts a relative tolerance parameter `_tol` (default `1e-6`) but was erroneously using the local variable `tol` when calling `PrincipalMoments`. This was causing inertia visuals to render improperly, with the `EquivalentBox` rotation not matching the principal moments. This includes new tests to cover the relative tolerance behavior and demonstrate the bug (build and run tests for https://github.com/ignitionrobotics/ign-math/commit/3776a65f991385ca150bce54913cdee8566a811e to confirm) and fixes it in https://github.com/ignitionrobotics/ign-math/commit/478ccf817424cc62c832ac6dac30d99ef0ad8303. The fix could have been just one character (`tol` -> `_tol`), but I moved the statement up above the definition of `tol` just to be safe.

## Checklist
- [X] Added tests
- [X] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [X] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))

### DIFF
--- a/src/MassMatrix3_TEST.cc
+++ b/src/MassMatrix3_TEST.cc
@@ -776,7 +776,8 @@ TEST(MassMatrix3dTest, EquivalentBox)
     EXPECT_EQ(m, m2);
   }
 
-  // box 1x4x9 rotated by 90 degrees around Z
+  // box 9x4x1 rotated by 90 degrees around Z
+  // equivalent to 4x9x1 after rotation
   {
     const double mass = 12.0;
     const math::Vector3d ixxyyzz(82, 17, 97);
@@ -786,13 +787,15 @@ TEST(MassMatrix3dTest, EquivalentBox)
     EXPECT_TRUE(m.EquivalentBox(size, rot, -1e-6));
     EXPECT_EQ(size, math::Vector3d(9, 4, 1));
     EXPECT_EQ(rot, math::Quaterniond(0, 0, IGN_PI/2));
+    // equivalent to 4x9x1 after rotation
+    EXPECT_EQ((rot * size).Abs(), math::Vector3d(4, 9, 1));
 
     math::MassMatrix3d m2;
     EXPECT_TRUE(m2.SetFromBox(mass, size, rot));
     EXPECT_EQ(m, m2);
   }
 
-  // box 1x4x9 rotated by 45 degrees around Z
+  // box 9x4x1 rotated by 45 degrees around Z
   {
     const double mass = 12.0;
     const math::Vector3d ixxyyzz(49.5, 49.5, 97);
--- a/src/MassMatrix3_TEST.cc
+++ b/src/MassMatrix3_TEST.cc
@@ -293,8 +293,12 @@ TEST(MassMatrix3dTest, PrincipalMoments)
     EXPECT_EQ(m.PrincipalMoments(), Ieigen);
     EXPECT_TRUE(m.IsPositive());
     EXPECT_FALSE(m.IsValid());
-    // Test with different tolerances
+    // Ratio between largest diagonal and non-diagonal elements is 0.5
+    // With relative tolerance of 0.49, the matrix is clearly non-diagonal,
+    // so expect PrincipalMoments to return the sorted eigenvectors.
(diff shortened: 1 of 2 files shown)


src/MassMatrix3_TEST.cc
```diff
@@ -305,6 +308,9 @@ TEST(MassMatrix3dTest, PrincipalMoments)
     EXPECT_EQ(m.PrincipalMoments(), Ieigen);
     EXPECT_TRUE(m.IsPositive());
     EXPECT_TRUE(m.IsValid());
+    // Test with different tolerances
+    EXPECT_EQ(m.PrincipalMoments(0.24), Ieigen);
+    EXPECT_EQ(m.PrincipalMoments(0.26), Ixxyyzz);
```
chapulina, line 313: These 2 numbers look very similar, is there a reason why both are interesting to test? It would be nice to mention on the comment. The same for the new lines above.

chapulina (approved):
Works for me, thanks for the quick fix!

chapulina (changes requested):
Actually no, wait, I'm checking something here

chapulina:
I was mistaken before, this doesn't fix the issue for me. I just tried your reproduction steps with this branch and it didn't work:

```python
from ignition.math import MassMatrix3d
from ignition.math import Vector3d, Quaterniond

mm3 = MassMatrix3d()
mm3.set_mass(605)
mm3.set_ixx(3219)
mm3.set_iyy(3219)
mm3.set_izz(7.28)
mm3.set_ixy(-0.43)
mm3.set_ixz(-2.56)
mm3.set_iyz(3.37)

size = Vector3d()
rot = Quaterniond()
mm3.equivalent_box(size, rot)

print(size)
```

The output is 

```
7.98598 0.284122 0.252113
```

But I'd expect `Z >> X and Z >> Y`

scpeters:
> But I'd expect `Z >> X and Z >> Y`

you need to consider the orientation `rot` as well. There are multiple equivalent combinations of box `size` and `rot` parameters. If the matrix is already diagonalized for the given tolerances, an identity quaternion is returned along with the original ordering of diagonal components. However, if it is not already diagonal, then the principal moments are [sorted from smallest to largest](https://github.com/ignitionrobotics/ign-math/blob/ignition-math6_6.10.0/include/ignition/math/MassMatrix3.hh#L693-L698) and the box orientation is computed accordingly.

src/MassMatrix3_TEST.cc
scpeters: I've expanded the comments in https://github.com/ignitionrobotics/ign-math/pull/424/commits/40fe3c471914cd7ecafe09362307e8e5b464a556; is it clearer now?

chapulina: Very helpful, thanks!

scpeters:
> > But I'd expect `Z >> X and Z >> Y`
> 
> you need to consider the orientation `rot` as well. There are multiple equivalent combinations of box `size` and `rot` parameters. If the matrix is already diagonalized for the given tolerances, an identity quaternion is returned along with the original ordering of diagonal components. However, if it is not already diagonal, then the principal moments are [sorted from smallest to largest](https://github.com/ignitionrobotics/ign-math/blob/ignition-math6_6.10.0/include/ignition/math/MassMatrix3.hh#L693-L698) and the box orientation is computed accordingly.

@chapulina I added an additional expectation to a test case for `EquivalentBox` in https://github.com/gazebosim/gz-math/pull/424/commits/31f43456bdc54d28732492c047f3db5f45e689c4 that should illustrate how the rotation affects the size vector

chapulina (approved):
Ok, this does solve the inertia visualization issue I had. I think that my workspace got messed up halfway and it looked like it didn't, but it did. 

Sorry for the noise and thanks for the quick fix!

PR #434, opened 2022-05-27
Compute distance between points on a general celestial body

adityapande-1995:

# 🎉 New feature

## Summary
The current ``SphericalCoordinates::Distance`` function uses hardcoded Earth's radius for its calculation. In order to compute distances between points for any non-Earth celestial body, we would need to consider its radius as well. This PR overloads the ``Distance()`` function to add an additional parameter, ``_radius``.

## Test it
Added a simple test case : Consider a hypothetical body with its radius half as that of the earth. The distance between 2 points on that body should be half as the distance between them on Earth. 

## Implementation variation
I could also implement radius as an optional parameter in the original ``SphericalCoordinates::Distance()`` function. 

Another idea is to keep a small database of radii of common celestial bodies, so "MOON" should automatically get you a radius of 1.7374 * 10^6 m. This way it makes sense to keep overloads of the ``Distance`` method, we can have a third one with the signature ``Distance(Angle, Angle, Angle, Angle, string _celestialBody)``.

## Checklist
- [x] Added tests


include/gz/math/SphericalCoordinates.hh
```diff
@@ -141,6 +141,24 @@ namespace gz
...
+      /// \param[in] _lonA Longitude of point A.
+      /// \param[in] _latB Latitude of point B.
+      /// \param[in] _lonB Longitude of point B.
+      /// \param[in] _radius Radius of the celestial body in metres.
+      /// \return Distance in meters.
+      public: static double Distance(const ignition::math::Angle &_latA,
+                                     const ignition::math::Angle &_lonA,
+                                     const ignition::math::Angle &_latB,
+                                     const ignition::math::Angle &_lonB,
+                                     const double &_radius);
```
chapulina, line 160: I think there are other functions which are currently hardcoded to the Earth and may eventually need to support other bodies.

The way that `SphericalCoordinates` exposes support for other bodies is through the `SurfaceType` enum. With that in mind, instead of having the user pass surface-specific parameters to each function, I think we should support other surface types. 

For your use case, I'd suggest either adding the surface type you need (maybe `MOON_SCS`) to the enum, or maybe adding a `CUSTOM` surface type. How I think those 2 options could work:

* Keep an internal `radius` variable, and initialize it to `g_EarthRadius` if the type is `EARTH_WGS84`, and do the same for other known types.
* Add a `double Radius() const` getter that returns that radius
* Add a `void SetRadius(double _radius)` setter to be used in case the surface type is `CUSTOM`
* Consider printing or returning errors if the user tries to call functions for `CUSTOM` without setting a radius first

```
auto moonSc = SphericalCoordinates(SurfaceType::MOON_SCS);
ASSERT_GT(moonSc.Radius(), 0);
ASSERT_GE(moonSc.Distance(latA, lonA, latB, lonB), 0);
```

```
auto moonSc = SphericalCoordinates(SurfaceType::CUSTOM);

// Radius not set for custom
ASSERT_LT(customSc.Radius(), 0);
ASSERT_LT(customSc.Distance(latA, lonA, latB, lonB), 0);;

// With custom radius
customSc.SetRadius(myRadius);

ASSERT_GT(customSc.Radius(), 0);
ASSERT_GE(customSc.Distance(latA, lonA, latB, lonB), 0);
```

chapulina (dismissed):

include/gz/math/SphericalCoordinates.hh
adityapande-1995: The method ``Distance()`` is currently a static method, so is it not bound to an instance. Setting the surface type will require a new non static method. Is that reasonable to add ?

adityapande-1995: Based on an offline discussion with @chapulina , I'll write a new non static method for calculating distance, that takes into account the surface type.

adityapande-1995:
TODO : Add API to set and get ellipsoid parameters for custom surface type, refactor tests.

adityapande-1995:
**TODO** : 
* Add more tests to make codecov bot happy.
* Add a warning or replace the static Distance function (?)

include/gz/math/SphericalCoordinates.hh
```diff
@@ -41,7 +41,15 @@ namespace gz
               {
                 /// \brief Model of reference ellipsoid for earth, based on
                 /// WGS 84 standard. see wikipedia: World_Geodetic_System
-                EARTH_WGS84 = 1
+                EARTH_WGS84 = 1,
+
+                /// |brief Model of the moon, based on the Selenographic
```
scpeters, line 46: replace `|` with `\` before `brief`

include/gz/math/SphericalCoordinates.hh
```diff
@@ -41,7 +41,15 @@ namespace gz
...
+                /// coordinate system, see wikipedia: Selenographic
+                /// Coordinate System.
+                MOON_SCS = 2,
+
+                /// |brief Custom surface type
```
scpeters, line 51: `\brief`

scpeters (commented):
I think we should deprecate the existing static `Distance` function and rename it to `DistanceWGS84` or something like that. We could add the same function alias to `ign-math6`

jennuine (commented):
Initial pass, +1 to deprecate the existing static `Distance()` function

include/gz/math/SphericalCoordinates.hh
adityapande-1995: Done here : https://github.com/gazebosim/gz-math/pull/434/commits/b401946aaf0b4b02c32fcada9cea585a48c4e40e

include/gz/math/SphericalCoordinates.hh
adityapande-1995: Done ehre : https://github.com/gazebosim/gz-math/pull/434/commits/b401946aaf0b4b02c32fcada9cea585a48c4e40e

include/gz/math/SphericalCoordinates.hh
```diff
@@ -73,6 +81,17 @@ namespace gz
       /// \param[in] _type SurfaceType specification.
       public: explicit SphericalCoordinates(const SurfaceType _type);
 
+      /// \brief Constructor with surface type input and properties
+      /// input. To be used for CUSTOM_SURFACE.
+      /// \param[in] _type SurfaceType specification.
```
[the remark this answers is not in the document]
adityapande-1995: Done here : https://github.com/gazebosim/gz-math/pull/434/commits/b401946aaf0b4b02c32fcada9cea585a48c4e40e

src/SphericalCoordinates.cc
```diff
@@ -38,12 +38,31 @@ const double g_EarthWGS84Flattening = 1.0/298.257223563;
...
+// Source : https://nssdc.gsfc.nasa.gov/planetary/factsheet/moonfact.html
+const double g_MoonFlattening = 0.0012;
+
 // Private data for the SphericalCoordinates class.
 class gz::math::SphericalCoordinates::Implementation
 {
   /// \brief Type of surface being used.
   public: SphericalCoordinates::SurfaceType surfaceType;
 
+  /// |brief Radius of the given SurfaceType.
```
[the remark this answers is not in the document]
adityapande-1995: Done

src/SphericalCoordinates.cc
```diff
@@ -125,6 +153,53 @@ SphericalCoordinates::SphericalCoordinates(const SurfaceType _type)
...
+  this->SetElevationReference(0.0);
+}
+
+//////////////////////////////////////////////////
+SphericalCoordinates::SphericalCoordinates(
+    const SurfaceType _type,
+    const double _radius,
+    const double _axisEquatorial,
+    const double _axisPolar,
+    const double _flattening)
```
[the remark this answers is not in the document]
adityapande-1995: Done here : https://github.com/gazebosim/gz-math/pull/434/commits/03869a999a85f1a5e71c707413d01655f993b40b

src/SphericalCoordinates.cc
```diff
@@ -197,6 +272,30 @@ void SphericalCoordinates::SetSurface(const SurfaceType &_type)
...
+      {
+      // Set the semi-major axis
+      this->dataPtr->ellA = g_MoonAxisEquatorial;
+
+      // Set the semi-minor axis
+      this->dataPtr->ellB = g_MoonAxisPolar;
+
+      // Set the flattening parameter
+      this->dataPtr->ellF = g_MoonFlattening;
+
```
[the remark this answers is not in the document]
adityapande-1995: Went with the first option, overloaded SetSurface here : https://github.com/gazebosim/gz-math/pull/434/commits/03869a999a85f1a5e71c707413d01655f993b40b

src/SphericalCoordinates.cc
```diff
@@ -125,6 +153,53 @@ SphericalCoordinates::SphericalCoordinates(const SurfaceType _type)
...
+    const double _axisPolar,
+    const double _flattening)
+  : SphericalCoordinates()
+{
+  if (_type != CUSTOM_SURFACE)
+  {
+    std::cerr << "Surface type should be set to SURFACE_CUSTOM"
+      << std::endl;
+    return;
+  }
```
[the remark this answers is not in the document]
adityapande-1995: Removed the check, now the user can supply different values for Earth/ Moon as well : https://github.com/gazebosim/gz-math/commit/03869a999a85f1a5e71c707413d01655f993b40b

adityapande-1995:
The failing codecov is around the deprecated ``Distance``` function. If I add a test for that, CI would fail due to warnings.

include/gz/math/SphericalCoordinates.hh
```diff
@@ -167,6 +240,18 @@ namespace gz
...
+      /// custom ellipsoid properties.
+      /// \param[in] _type SurfaceType value.
+      /// \param[in] _axisEquatorial Equatorial axis of the surface.
+      /// \param[in] _axisPolar Polar axis of the surface.
+      /// \param[in] _flattening Falttening parameter of the surface.
+      public: void SetSurface(
+                  const SurfaceType &_type,
+                  const double _axisEquatorial,
+                  const double _axisPolar,
+                  const double _flattening);
```
jennuine, line 253: Shouldn't we have the `_radius` as a parameter as well?

include/gz/math/SphericalCoordinates.hh
```diff
@@ -136,15 +158,66 @@ namespace gz
       /// \param[in] _latB Latitude of point B.
       /// \param[in] _lonB Longitude of point B.
       /// \return Distance in meters.
-      public: static double Distance(const gz::math::Angle &_latA,
+      /// \deprecated Use DistanceWGS84 instead.
+      public: GZ_DEPRECATED(7) static double Distance(
```
jennuine, line 162: Can you add this to the [Migration.md](https://github.com/gazebosim/gz-math/blob/ign-math6/Migration.md)?

include/gz/math/SphericalCoordinates.hh
```diff
@@ -136,15 +158,66 @@ namespace gz
...
+                  const gz::math::Angle &_latB,
+                  const gz::math::Angle &_lonB);
+
       /// \brief Get SurfaceType currently in use.
       /// \return Current SurfaceType value.
       public: SurfaceType Surface() const;
 
+      /// \brief Get the radius of the surface.
+      /// \return radius of the surface in use.
+      public: double GetSurfaceRadius();
```
jennuine, line 207:
```suggestion
      public: double SurfaceRadius();
```
We should remove the `Get` from this method and the methods below based on our [style guide](https://gazebosim.org/docs/all/contributing#style-guides) # 15

src/SphericalCoordinates.cc
```diff
@@ -125,6 +153,48 @@ SphericalCoordinates::SphericalCoordinates(const SurfaceType _type)
...
+      _axisPolar, _flattening);
+
+  if (_radius > 0)
+  {
+    this->dataPtr->surfaceRadius = _radius;
+  }
+  else
+  {
+    std::cerr << "Value of _radius should be greater than zero "
+      " defaulting to Earth's flattening value."<< std::endl;
```
jennuine, line 194:
```suggestion
      " defaulting to Earth's radius value."<< std::endl;
```

src/SphericalCoordinates.cc
```diff
@@ -219,6 +313,63 @@ void SphericalCoordinates::SetSurface(const SurfaceType &_type)
...
+  }
+
+  if (_flattening >= 0)
+  {
+    this->dataPtr->ellF = _flattening;
+  }
+  else
+  {
+    std::cerr << "Value of _flattening should be greater than zero "
+      " defaulting to Earth's flattening value."<< std::endl;
```
jennuine, line 362: Just checking, the condition has `_flattening >= 0`, should the error message say `.. greater or equal to zero..`?

adityapande-1995: Done here : https://github.com/gazebosim/gz-math/pull/434/commits/b1baee53f003c614bfeec7c28afb2458a82cd7bc

src/SphericalCoordinates.cc
adityapande-1995: Done here : https://github.com/gazebosim/gz-math/pull/434/commits/b1baee53f003c614bfeec7c28afb2458a82cd7bc

include/gz/math/SphericalCoordinates.hh
adityapande-1995: Done here : https://github.com/gazebosim/gz-math/pull/434/commits/b1baee53f003c614bfeec7c28afb2458a82cd7bc

include/gz/math/SphericalCoordinates.hh
adityapande-1995: Done here : https://github.com/gazebosim/gz-math/pull/434/commits/b1baee53f003c614bfeec7c28afb2458a82cd7bc

include/gz/math/SphericalCoordinates.hh
adityapande-1995: Done here : https://github.com/gazebosim/gz-math/pull/434/commits/765b8d9286ae44a856a61269e5c28bc159e3589b

src/SphericalCoordinates.cc
```diff
@@ -208,6 +251,36 @@ void SphericalCoordinates::SetSurface(const SurfaceType &_type)
...
+
+      // Set the second eccentricity ellipse parameter
+      // https://en.wikipedia.org/wiki/Eccentricity_(mathematics)#Ellipses
+      this->dataPtr->ellP = sqrt(
+          std::pow(this->dataPtr->ellA, 2) / std::pow(this->dataPtr->ellB, 2) -
+          1.0);
+
+      // Set the radius of the surface.
+      this->dataPtr->surfaceRadius = g_MoonRadius;
+
```
jennuine, line 283: We should check for `CUSTOM_SURFACE` and print a warning/error to use `SetSurface(type, radius, axisEquatorial, axisPolar, flattening)` instead

src/SphericalCoordinates_TEST.cc
```diff
@@ -114,6 +129,54 @@ TEST(SphericalCoordinatesTest, SetFunctions)
     EXPECT_EQ(sc.HeadingOffset(), heading);
     EXPECT_NEAR(sc.ElevationReference(), elev, 1e-6);
   }
+
+  // Moon surface type
+  st = math::SphericalCoordinates::MOON_SCS;
+  math::SphericalCoordinates moonSC(st);
+  moonSC.SetSurface(st);
+  EXPECT_EQ(moonSC.Surface(), st);
```
jennuine, line 137: We should test the other new get functions (e.g., `SurfaceRadius()`, `SurfaceAxisEquatorial()`, etc). First for the default parameters above (i.e., matches Earth values) then for the Moon surface type here.

jennuine (commented):
LGTM, left some minor comments. I'm new to `gz-math` so this could use a final review from @scpeters

src/SphericalCoordinates.cc
adityapande-1995: Done

include/gz/math/SphericalCoordinates.hh
```diff
@@ -167,6 +240,19 @@ namespace gz
...
 
+      /// \brief Set SurfaceType for planetary surface model with
+      /// custom ellipsoid properties.
+      /// \param[in] _type SurfaceType value.
+      /// \param[in] _axisEquatorial Equatorial axis of the surface.
+      /// \param[in] _axisPolar Polar axis of the surface.
+      /// \param[in] _flattening Falttening parameter of the surface.
+      public: void SetSurface(
+                  const SurfaceType &_type,
+                  const double _radius,
```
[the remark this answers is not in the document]
adityapande-1995: Done

ahcorde (dismissed):
I saw that you added the Python interface, do you mind to add the tests too ? otherwise can you add an issue ?

include/gz/math/SphericalCoordinates.hh
```diff
@@ -136,15 +158,66 @@ namespace gz
...
+      /// latitude and longitude. It assumes that both points are at sea level.
+      /// Example: _latA = 38.0016667 and _lonA = -123.0016667) represents
+      /// the point with latitude 38d 0'6.00"N and longitude 123d 0'6.00"W.
+      /// This is different from the static Distance() method as it takes into
+      /// account the surface type.
+      /// \param[in] _latA Latitude of point A.
+      /// \param[in] _lonA Longitude of point A.
+      /// \param[in] _latB Latitude of point B.
+      /// \param[in] _lonB Longitude of point B.
+      /// \return Distance in meters.
```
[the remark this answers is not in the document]
scpeters: we use meters everywhere in our API, so I think we can just leave it at that

include/gz/math/SphericalCoordinates.hh
```diff
@@ -73,6 +81,21 @@ namespace gz
...
+      /// \brief Constructor with surface type input and properties
+      /// input. To be used for CUSTOM_SURFACE.
+      /// \param[in] _type SurfaceType specification.
+      /// \param[in] _radius Radius of the surface.
+      /// \param[in] _axisEquatorial Semi major axis of the surface.
+      /// \param[in] _axisPolar Semi minor axis of the surface.
+      /// \param[in] _axisPolar Flattening parameter of the surface.
+      public: SphericalCoordinates(
+            const SurfaceType _type,
+            const double _radius,
```
scpeters, line 93: this radius value is used as an approximate average radius for computations based on a spherical surface. I would recommend computing this radius value as the arithmetic mean of the equatorial and polar axes rather than requiring users to specify it

* see https://en.wikipedia.org/wiki/Earth_radius#Arithmetic_mean_radius

`(2 * _axisEquatorial + _axisPolar) / 3`

include/gz/math/SphericalCoordinates.hh
```diff
@@ -73,6 +81,21 @@ namespace gz
...
+            const double _axisEquatorial,
+            const double _axisPolar,
+            const double _flattening
```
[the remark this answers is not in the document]
adityapande-1995: Done here : https://github.com/gazebosim/gz-math/pull/434/commits/e0525a6c09647dfbfea54c2287d2fcf522527315

include/gz/math/SphericalCoordinates.hh
adityapande-1995: Done https://github.com/gazebosim/gz-math/pull/434/commits/b8c488711b0b2b15d3283019ff937d9ebcd7886e

src/SphericalCoordinates.cc
```diff
@@ -208,6 +249,42 @@ void SphericalCoordinates::SetSurface(const SurfaceType &_type)
...
+      // Set the first eccentricity ellipse parameter
+      // https://en.wikipedia.org/wiki/Eccentricity_(mathematics)#Ellipses
+      this->dataPtr->ellE = sqrt(1.0 -
+          std::pow(this->dataPtr->ellB, 2) / std::pow(this->dataPtr->ellA, 2));
+
+      // Set the second eccentricity ellipse parameter
+      // https://en.wikipedia.org/wiki/Eccentricity_(mathematics)#Ellipses
+      this->dataPtr->ellP = sqrt(
+          std::pow(this->dataPtr->ellA, 2) / std::pow(this->dataPtr->ellB, 2) -
+          1.0);
```
scpeters, line 277: nit: the `ellE` and `ellP` computations are repeated between this `MOON_SCS` case and the prior `EARTH_WGS84` case. We could potentially move them outside of this `switch` block if the `CUSTOM_SERVICE` and `default` cases had a `return` instead of `break`, but this isn't a big deal, so it's not required

scpeters (approved):

PR #435, opened 2022-05-31
Update GoogleTest to latest version

mjcarroll:
Import latest version of GoogleTest from upstream and adjust build accordingly.


ahcorde (approved):

chapulina:
Would it be possible to target this at `ign-math6`? I'm working on some tests that only pass with these changes 😬

mjcarroll:
> I'm working on some tests that only pass with these changes.

Likely yes, but what kind of tests out of curiosity?

PR #437, opened 2022-06-06
ign -> gz Macro Migration : gz-math

methylDragon:
See: https://github.com/gazebo-tooling/release-tools/issues/737


src/python_pybind11/src/Helpers.cc
```diff
@@ -164,23 +164,23 @@ void defineMathHelpers(py::module &m)
    .def("parse_float",
         &gz::math::parseFloat,
         "parse string into an float")
-   .def("ign_sphere_volume",
+   .def("gz_sphere_volume",
```
chapulina, line 167: There are not being tick-tocked, right? Can we have both the old and new versions? Not sure how trivial it would be to add warnings.

chapulina (changes requested):
Since this will break downstream and will need to be merged in sync, how about also tackling the remaining `IGN_` variables in this PR? I think it's mainly the `IGN_*_SIZE_T`s in `Helpers.hh`

src/python_pybind11/src/Helpers.cc
methylDragon: This should do it, I think!

methylDragon:
> Since this will break downstream and will need to be merged in sync, how about also tackling the remaining `IGN_` variables in this PR? I think it's mainly the `IGN_*_SIZE_T`s in `Helpers.hh`

I was thinking of doing that, but I'm not sure how we want to do it.

```
static const size_t GZ_ZERO_SIZE_T  = 0u;
constexpr auto IGN_ZERO_SIZE_T = &GZ_ZERO_SIZE_T;
```

?

methylDragon:
Tick-tocked! https://github.com/gazebosim/gz-math/pull/437/commits/1a32dfcf876db197f36aa8b1d59c9c793dffc9d5

chapulina (approved):
🚀

chapulina (approved):

methylDragon:
@osrf-jenkins run tests please!

Issue #144, opened 2020-07-28
Add variable-sized Matrix


mcres:
While working on [this PR](https://github.com/ignitionrobotics/ign-gazebo/pull/229), @mabelzhang and I thought that it would be cool to have something like Eigen's `MatrixXd`. Ignition Math already has the [Matrix3](https://ignitionrobotics.org/api/math/6.4/classignition_1_1math_1_1Matrix3.html) and [Matrix4](https://ignitionrobotics.org/api/math/6.4/classignition_1_1math_1_1Matrix4.html) classes, which are 3x3 and 4x4, respectively.

However, I think it would be a good idea being able to create a custom-sized Matrix in order to make some basic computations on image processing. For this, this new class could be integrated into [ignition::common::Image](https://ignitionrobotics.org/api/common/3.5/classignition_1_1common_1_1Image.html).

Plus, I believe it would be useful for a general user as well. It would allow the user to quickly create, modify and compute matrices on whatever it is her/him is working on. I think that the goal is being able to do some basic linear algebra for entry-level users and basic applications so the user doesn't have to learn `Eigen`. Of course, for more complicated things, the user should be encouraged to use `Eigen`, as it is more mature in this field and it has a lot of functionality.

PR #442, opened 2022-06-17
Add MatrixX class

chapulina:
# 🎉 New feature

* Addresses https://github.com/gazebosim/gz-math/issues/144

## Summary

We're coming for you, Eigen! (Well, that's what a lot of people think about `gz-math`... 😆 )

We need a `Matrix6d` to implement

* https://github.com/gazebosim/gz-sim/issues/1462

Instead of copying the existing `Matrix3` and `Matrix4` classes, I thought it would be useful to implement a matrix class that can have any size. The size is defined at compile time and can't be changed for a given object, so the matrix can't be dynamically resized.

I only implemented very basic functionality, more functions can be added to this class as needed. I think we probably have everything that's needed to implement added mass, but if we don't I can add new functions later.

I experimented with different `std` algorithms to substitute the nested for loops, but they were all too confusing to use with nested `std::array`. I'm open to suggestions though.

### TODO

- [ ] I'd like to get some high-level feedback on this PR and the added APIs before adding Python bindings for this class.

## Test it

Take a look at the added tests and run the new example.

## Checklist
- [x] Added tests
- [x] ~~Updated migration guide (as needed)~~

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -15,6 +15,9 @@ target_link_libraries(diff_drive_odometry ignition-math${IGN_MATH_VER}::ignition
 add_executable(gauss_markov_process gauss_markov_process_example.cc)
 target_link_libraries(gauss_markov_process ignition-math${IGN_MATH_VER}::ignition-math${IGN_MATH_VER})
 
+add_executable(matrixx_example matrixx_example.cc)
+target_link_libraries(matrixx_example ignition-math${IGN_MATH_VER}::ignition-math${IGN_MATH_VER})
+
 add_executable(vector2_example vector2_example.cc)
 target_link_libraries(vector2_example ignition-math${IGN_MATH_VER}::ignition-math${IGN_MATH_VER})
 
--- a/examples/matrixx_example.cc
+++ b/examples/matrixx_example.cc
@@ -19,21 +19,21 @@
 
 int main(int argc, char **argv)
 {
-  ignition::math::MatrixX<double, 2, 3> matd23A(
+  ignition::math::MatrixXd<2, 3> matA(
       0.1, 0.2, 0.3,
       0.4, 0.5, 0.6);
+  std::cout << "Matrix A: " << matA << std::endl;
 
-  ignition::math::MatrixX<double, 2, 3> matd23B(
+  ignition::math::MatrixXd<2, 3> matB(
       1.1, 1.2, 1.3,
       1.4, 1.5, 1.6);
+  std::cout << "Matrix B: " << matB << std::endl;
 
-  auto matSum = matd23A + matd23B;
+  auto matSum = matA + matB;
+  std::cout << "A + B sum: " << matSum << std::endl;
 
-  std::cout << matd23A << std::endl;
-
-  std::cout << matd23B << std::endl;
-
-  std::cout << matSum << std::endl;
+  auto matATrans = matA.Transposed();
+  std::cout << "A transposed: " << matATrans << std::endl;
 
   return 0;
 }
--- a/examples/matrixx_example.cc
+++ b/examples/matrixx_example.cc
@@ -1,0 +1,39 @@
+/*
+ * Copyright (C) 2022 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+#include <iostream>
+#include <ignition/math/MatrixX.hh>
+
+int main(int argc, char **argv)
+{
+  ignition::math::MatrixX<double, 2, 3> matd23A(
+      0.1, 0.2, 0.3,
+      0.4, 0.5, 0.6);
+
+  ignition::math::MatrixX<double, 2, 3> matd23B(
+      1.1, 1.2, 1.3,
+      1.4, 1.5, 1.6);
+
+  auto matSum = matd23A + matd23B;
+
+  std::cout << matd23A << std::endl;
+
+  std::cout << matd23B << std::endl;
+
+  std::cout << matSum << std::endl;
+
+  return 0;
+}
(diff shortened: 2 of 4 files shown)


scpeters:
I'm leaning against `MatrixXd` to be honest. I think it's not strictly necessary for hydrodynamics, and I'd rather not add to the maintenance burden of gz-math by trying to tackle Eigen's functionality.

I think we could compose a `gz::math::HydrodynamicAddedMass` class using  a pair of `MassMatrix3`s for the symmetric, block-diagonal 3x3 matrices and a `Matrix3` for the upper / lower block diagonal component. Defining the class using sub-matrices could also help avoid confusion about the ordering of indices (rotation vs position) by ensuring that proper names are used.

mjcarroll:
I'm inclined to agree with @scpeters on reproducing Eigen functionality. In this case, `gz-sim` already has a dependency on `libgz-math-eigen3`, so Eigen can be used there without introducing a new dependency.

My main concern would be that we are missing out on a lot of testing/optimizations that Eigen has already gone through for multiple platforms (hardware and OS).

chapulina:
> I think we could compose a gz::math::HydrodynamicAddedMass class 

That feels a bit too specific to me. Hydrodynamic added mass is one use case for exposing all the fields of a spatial mass matrix, but I can imagine other advanced users may want to directly tweak those fields as well. [Drake](https://drake.mit.edu/doxygen_cxx/group__multibody__spatial__inertia.html) and [DART](https://dartsim.github.io/dart/main/d6/d91/classdart_1_1dynamics_1_1Inertia.html#aa4661f9950d5958efa4c8609d42aedf7) are just two examples of engines that expose the full spatial tensor without making it specific to hydrodynamics.

If a `Matrix6` class is not desired, I'd suggest we just add more `Matrix3` fields directly to `Inertial` to hold various parts of the matrix - but that feels to me more like a workaround our limited API than a careful design choice.

> gz-sim already has a dependency on libgz-math-eigen3

Yeah but SDFormat doesn't, and neither does the `math::Inertial` class. I can hear an argument in favor of deprecating our existing vector and matrix classes completely in favor of Eigen, but that would be a longer-term task.

> we are missing out on a lot of testing/optimizations that Eigen has already gone through for multiple platforms (hardware and OS).

Yeah definitely, and that's also true for all our existing vector and matrix classes.

> against MatrixXd to be honest. I think it's not strictly necessary for hydrodynamics, and I'd rather not add to the maintenance burden of gz-math by trying to tackle Eigen's functionality.

I hear you. But to summarize the comments above, the choices I see here are:

1. Composing a 6x6 matrix out of 3x3 sub-matrices - which feels like a workaround for a limited API
2. Doing things the "Gazebo Math way" and adding a `Matrix6` class similar to `Matrix3` and `Matrix4` - I'm not opposed to that, I just thought we could try to be more flexible and reduce duplication - which in itself should be better for maintenance.
3. Going all in on Eigen - which is not doable on Fortress, and would require more effort to make our API consistent - but I'm not opposed to if we get the funds to do it right
4. Going with `MatrixX` - I saw issue #144 and thought this would be an opportunity to tackle it. I don't think we should try to implement various algorithms for this class, I think it could just be handy to hold data, which is what we need for added mass.

mjcarroll:
>  I think it could just be handy to hold data, which is what we need for added mass.

I think in this case, having the Matrix is useful.  I think, in general, it would be nice if people were using Eigen when possible to actually do computations/manipulations.  I understand that we may still need some of our own datatypes for convenience in terms of serialization and interactions with other locations in our APIs, but if we can encourage "jumping out" to Eigen for math heavy operations, I think that would be the ideal mix.

chapulina:
> if we can encourage "jumping out" to Eigen for math heavy operations, I think that would be the ideal mix.

I believe we've been doing this higher up the stack. `gz-physics` makes heavy use of Eigen, as well as various `gz-sim` plugins, some `gz-rendering` functionality... Maybe I shouldn't have joked about Eigen on the PR description, it definitely has a place in our development, but I don't think it's a good match in this specific case.

> the choices I see here are

So for completeness, I'll mention my preference would be to go with option 4, but I could compromise for 2. I don't think 1 is a good option and 3 isn't feasible at this moment. @scpeters , please let me know what you think.

scpeters:
> > I would propose adding this class to the eigen3 component.
> 
> Are you proposing also moving the existing `Inertial` and `MassMatrix3` classes to the eigen component? The idea of the [Hydrodynamic Added Mass Proposal](http://sdformat.org/tutorials?tut=added_mass_proposal&branch=chapulina/added_mass) is to augment the existing class, not to create a separate implementation.

Sorry I didn't realize you were proposing to modify the existing `math::Inertial` class to included added mass. Won't that break ABI on `ign-math6`?

chapulina:
> Won't that break ABI on ign-math6?

Ouch yeah I forgot that these classes aren't PIMPL'ed. I guess I'll have to target this at Garden then.

chapulina:
> I still think there is a place for gz::math::Vector[234] and gz::math::Matrix[34]. I don't have any immediate plans to deprecate them and switch fully to Eigen.

Could you elaborate on that? What's the place for these classes when we're using Eigen throughout our stack? And why should we maintain matrix 3 and 4, but not 6 or X?

chapulina:
> I think we could compose a gz::math::HydrodynamicAddedMass class using a pair of MassMatrix3s for the symmetric, block-diagonal 3x3 matrices and a Matrix3 for the upper / lower block diagonal component. Defining the class using sub-matrices could also help avoid confusion about the ordering of indices (rotation vs position) by ensuring that proper names are used.

> avoiding the indexing confusion

After giving this point a bit more consideration, I still think that this would be clunky. The `MassMatrix3` class doesn't seem fit to be used this way - do we just leave the `mass` invalid? What do `DiagonalMoments` mean in this case? Instead of using proper names, we'd be giving a new meaning to different concepts. That's more confusing.

What I think you really want is a more generic `SymmetricMatrix3` class. And at this point, we could go directly for a `SymmetricMatrix6` or `SymmetricMatrixX` class that would address our use case directly. I started a draft in that direction in https://github.com/gazebosim/gz-math/pull/447.

> it doesn't actually store a 3x3 matrix. Instead, it stores the mass, moments of inertia, and products of inertia separately. You can get a 3x3 matrix out of it, but that's not how it's stored.

I get the point that a generic matrix is not the optimal storage for a symmetric matrix. I started an implementation on #447 which is called `FluidAddedMass`, but it's really a generic `SymmetricMatrix6`. 

Even with that, I still feel that a `Matrix6` is missing. At some point we'll need to construct a 6x6 spatial matrix from the current terms in `math::Inertial`, and add that to the added mass matrix. Without a generic `Matrix6` representation the implementation will probably just add term by term and assign them to the physics engine, which feels a bit messy, but I can live with.

scpeters:
since we won't be using Eigen for this; I think I would favor a `Matrix6` class over `MatrixX`. I would also suggest helper functions for getting and setting 3x3 subsets of the Matrix6 from a Matrix3 (block diagonal upper left and lower right, and block off-diagonal upper right and lower left). This would make it easy enough to map Inertial values to a spatial inertia matrix. I don't think we need to support symmetric matrices at the class level

chapulina:
Thanks, @scpeters , I'll open a separate PR with a `Matrix6` implementation.

PR #447, opened 2022-06-29
Fluid added mass / symmetric matrix

chapulina:
# 🎉 New feature

* Alternative to https://github.com/gazebosim/gz-math/pull/442

## Summary

Based on the feedback on #442, I started implementing a specific class to hold fluid added mass. I stopped halfway through the implementation because various aspects of this didn't feel right to me, so I'm opening this for others to take a look and we have something more concrete to discuss.

I couldn't come up with ways to make the API a bit more semantically specific to added mass, and I don't even know if we want to. Maritime literature usually uses numeric indices to denote these terms within the matrix and AFAICT it is not common to decompose the matrix. It's true that we may need to shift submatrices around to adapt to different conventions, such as swapped rotational and translational components, or even shift things even more if a physics engine doesn't use Z-up. But I feel like this is the kind of thing that we can address with documentation, not APIs.

So my inclination at this point is to go with either:

* The `MatrixX` class in #442 
* This PR, with `FluidAddedMass` renamed to `SymmetricMatrixX` and made general since there isn't anything specific to "6" in the implementation so far.

## Checklist
- [ ] Added tests

### DIFF
--- a/include/gz/math/FluidAddedMass.hh
+++ b/include/gz/math/FluidAddedMass.hh
@@ -1,0 +1,113 @@
+/*
+ * Copyright (C) 2022 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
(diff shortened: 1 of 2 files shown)


chapulina:
> a separate Inertial class that uses it

Can you elaborate the motivation? What's the downside of having the added mass in the existing inertial class? The added mass doesn't affect the rigid body terms unless it is set, so it is a backwards-compatible addition. How do you expect the two inertial classes to interact? Why have two classes that duplicate inertial information?

Are you proposing we also separate it on the SDF spec? It sounded like you were ok keeping it within `<inertial>` on https://github.com/gazebosim/sdf_tutorials/pull/76. So it feels odd to have it all working together on SDF, but not on Gazebo Math.

And how do you expect downstream libs like `gz-physics` and `gz-sim` to handle these 2 different inertial classes?

chapulina:
Per https://github.com/gazebosim/gz-math/pull/442#issuecomment-1175427987, I'll be opening a new PR with a `Matrix6` class that follows the `Matrix[3|4]` pattern.

PR #456, opened 2022-07-08
Adds a Time Interpolation API

arjo129:
# 🎉 New feature

Closes #<NUMBER>

## Summary
This PR adds a `TimeVaryingVolumetricGrid`. The core idea is this extends `VolumetricGrid` by stepping through time. It uses template speciallization to leave the room open to custom implementations of the index.

## Checklist
- [x] Added tests


ahcorde (commented):
If you don't have the time to generate the Python interface. Do you mind to open a issue ?

hidmic (commented):
@arjo129 First, high-level pass. As I try to visualize how everything comes together, a few pain points start creeping in.

include/gz/math/TimeVaryingVolumetricGridLookupField.hh
```diff
@@ -0,0 +1,217 @@
...
+        while (nextTime != this->gridFields.end()
+          && nextTime->first <= _time)
+        {
+          newSess.iter = nextTime;
+          nextTime = std::next(nextTime);
+        }
+        return newSess;
+      }
+
+      public: std::pair<InterpolationPoint4D<T, V>, InterpolationPoint4D<T, V>>
```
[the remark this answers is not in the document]
arjo129: Fair enough we could just add another `AxisIndex` to `VolumetricGridFieldData`. This does have slight performance advantage for this usecase though.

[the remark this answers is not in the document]
hidmic: > This does have slight performance advantage for this usecase though.

In what sense?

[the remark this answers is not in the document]
arjo129: By keeping track of the session iter we can use `std::next` which has amortized O(1) (although worst case O(logn)) IIRC. That being said I agree I can simplify this by just reusing AxisIndex and save us this headache.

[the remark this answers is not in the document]
hidmic: Right, but N-dimensional arrays will outperform hash maps and binary trees if accessed in a cache conscious way.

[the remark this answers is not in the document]
arjo129: So I had originally proposed N-d arrays to Ben and mabel however, given precision conversion errors between UTM<=> and Lat/Lon, the grids are not quite consistent. Ben was the one who actually told u to use trees as thats what works for mbari. W.r.t to time we could probably get away assuming our data is well ordered and use an `std::vector` instead of std::map then. Arbitrary lookup would still rely on a binary search using `std::lower_bound`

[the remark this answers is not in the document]
hidmic: @arjo129 ahh, I see now. Having a session/cursor for the index makes total sense.

src/TimeVaryingVolumetricGridLookupField_TEST.cc
```diff
@@ -0,0 +1,90 @@
...
+  timeVaryingField.AddVolumetricGridField(1, scalarIndex);
+
+  {
+    // Get data at T=0
+    auto session = timeVaryingField.CreateSession();
+    auto points = timeVaryingField.LookUp(session, 0, Vector3d{0.5, 0.5, 0.5});
+    auto res = timeVaryingField.EstimateQuadrilinear<double>(
+      session,
+      points,
+      valuesTime0,
```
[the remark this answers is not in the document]
arjo129: See 3ae3d0a

include/gz/math/TimeVaryingVolumetricGrid.hh
```diff
@@ -0,0 +1,114 @@
...
+
+  public: std::optional<S> StepTo(const S &_session, const T &_time) const
+  {
+
+  }
+
+  public: std::optional<V> LookUp(const S &_session, const Vector3d &_pos) const
+  {
+
+  }
```
hidmic, line 42: @arjo129 nit: there's no need to define empty implementations for methods that won't ever be invoked (nor should they be)

include/gz/math/TimeVaryingVolumetricGrid.hh
```diff
@@ -0,0 +1,114 @@
...
+      _session,
+      points,
+      values,
+      values,
+      _pos);
+    return result;
+  }
+
+  std::vector<V> values;
+  TimeVaryingVolumetricGridLookupField<T, V, InMemorySession<T,double>> indices;
```
hidmic, line 74: @arjo129 hmm, you want these to be private?

include/gz/math/TimeVaryingVolumetricGrid.hh
```diff
@@ -0,0 +1,114 @@
...
+    return grid;
+  }
+
+  private: std::map<T, std::vector<std::pair<Vector3d, V>>> _points;
+};
+
+}
+}
+
+#endif
```
hidmic, line 114: @arjo129 nit: missing newline

include/gz/math/TimeVaryingVolumetricGridLookupField.hh
```diff
@@ -0,0 +1,231 @@
...
+      }
+
+      public: std::optional<InMemorySession<T, V>> StepTo(
+        const InMemorySession<T, V> &_session, const T &_time) const {
+        if (_session.iter == gridFields.end())
+        {
+          return std::nullopt;
+        }
+
+        InMemorySession<T, V> newSess(_session);
```
hidmic, line 129: @arjo129 nit: can we avoid a session copy if not necessary? User code will step frequently, as it can't know (nor should it have to know) when a step will actually have an effect.

include/gz/math/VolumetricGridLookupField.hh
```diff
@@ -143,6 +143,53 @@ namespace gz
...
+        {
+          assert(_indices.size() == _cloud.size());
+          // NOTE: This part of the code assumes an exact grid of points.
+          // The grid may be distorted or the stride between different points
+          // may not be the same, but fundamentally the data is structured still
+          // forms a grid-like structure. It keeps track of the axis indices for
+          // each point in the grid.
+          for(auto pt : _cloud)
+          {
+            x_indices_by_lat.AddIndexIfNotFound(pt.X());
```
hidmic, line 161: @arjo129 nit^2, preexisting: lat, lon, depth imply a given set of coordinates that is not required by the API itself (it could easily be cartesian)

include/gz/math/TimeVaryingVolumetricGridLookupField.hh
arjo129: So something like `bool StepTo(InMemorySession<T, V> &_session, const T &_time) const` instead and modify the iterator in place?

hidmic: That's an option. The alternative would be to encapsulate all access through a mutable session (ie. cursor) and step the session (ie. cursor) itself.

include/gz/math/TimeVaryingVolumetricGrid.hh
arjo129: Thanks for spotting these.

include/gz/math/TimeVaryingVolumetricGrid.hh
```diff
@@ -0,0 +1,113 @@
...
+namespace math
+{
+template<typename T, typename V, typename S>
+class TimeVaryingVolumetricGrid
+{
+  public: S CreateSession() const;
+
+  public: std::optional<S> StepTo(const S &_session, const T &_time) const;
+
+  public: std::optional<V> LookUp(const S &_session, const Vector3d &_pos)
```
hidmic, line 39: @arjo129 meta: why does `time` support a generic scalar type but `pos` is forced to `double`?

include/gz/math/TimeVaryingVolumetricGrid.hh
```diff
@@ -0,0 +1,113 @@
...
+  public: void AddPoint(
+    const T &_time, const Vector3d &_position, const V &_value)
+  {
+    _points[_time].emplace_back(_position, _value);
+  }
+
+  public:
+  TimeVaryingVolumetricGrid<T, V, InMemorySession<T, double>> Build() const
+  {
+    TimeVaryingVolumetricGrid<T, V, InMemorySession<T, double>> grid;
```
hidmic, line 90: @arjo129 nit: may be worth adding an `InMemoryTimeVayringVolumetricGrid` alias for convenience.

include/gz/math/VolumetricGridLookupField.hh
```diff
@@ -47,7 +47,7 @@ namespace gz
         private: AxisIndex<T> y_indices_by_lon;
 
         private: std::vector<
-          std::vector<std::vector<std::optional<std::size_t>>>> index_table;
+          std::vector<std::vector<std::optional<I>>>> index_table;
```
scpeters, line 50: it looks like this is the only breaking change in this PR, so we can split this out to merge before feature freeze if the rest of the changes have not been reviewed

arjo129: See #475

include/gz/math/TimeVaryingVolumetricGrid.hh
```diff
@@ -0,0 +1,152 @@
...
+/// specialization that loads all data into memory. To use this class you should
+/// use `CreateSession` so you can step through time.
+template<typename T, typename V, typename S, typename P>
+class TimeVaryingVolumetricGrid
+{
+  /// \brief Creates a session. Call this before querying the interface.
+  public: S CreateSession() const;
+
+  /// \brief Steps the session to a fixed time step
+  /// \param _session - The session to be stepped
```
hidmic, line 45: @arjo129 nit: 
```suggestion
  /// \param[in] _session The session to be stepped
```
Same everywhere else.

include/gz/math/TimeVaryingVolumetricGrid.hh
```diff
@@ -0,0 +1,152 @@
...
+  /// \param _time - The time at which the session should be set.
+  /// Returns the session cursor if the step was successful. If there was
```
hidmic, line 47: @arjo129 nit: consider using `\return`. Same elsewhere.

include/gz/math/TimeVaryingVolumetricGrid.hh
```diff
@@ -0,0 +1,152 @@
...
+  /// no more data return nullopt.
+  public: std::optional<S> StepTo(const S &_session, const T &_time) const;
+
+  /// \brief Looks up a given point in time.
```
hidmic, line 51: @arjo129 nit: 
```suggestion
  /// \brief Looks up a value at a given point in time.
```

include/gz/math/TimeVaryingVolumetricGridLookupField.hh
```diff
@@ -0,0 +1,291 @@
...
+#include <optional>
+#include <utility>
+#include <vector>
+
+namespace gz
+{
+  namespace math
+  {
+    /// \brief Lookup table for a time-varying volumetric dataset.
+    /// This is an unimplimented template as the actual methods depend on the
```
hidmic, line 33: @arjo129 typo: 
```suggestion
    /// This is an unimplemented template as the actual methods depend on the
```

include/gz/math/TimeVaryingVolumetricGridLookupField.hh
```diff
@@ -0,0 +1,291 @@
...
+        }
+        newSess.time = _time;
+        return newSess;
+      }
+
+      /// \brief Documentation inherited
+      public: std::vector<InterpolationPoint4D<T, V>>
+        LookUp(const InMemorySession<T, V> &_session,
+          const Vector3<V> &_point,
+          const Vector3<V> &_tol = Vector3<V>{0.5, 0.5, 0.5}) const {
```
hidmic, line 170: @arjo129 meta: depending on the scale and nature of `V` values, `0.5` can be huge for a tolerance. I don't mind strongly about the arbitrary value, but I do think there should be a way to change this tolerance when using a `TimeVaryingVolumetricGrid`.

include/gz/math/TimeVaryingVolumetricGridLookupField.hh
```diff
@@ -0,0 +1,291 @@
...
+          _position,
+          _values2,
+          _default
+        );
+
+        if (res1.has_value() || res2.has_value())
+        {
+          InterpolationPoint1D<T>
+            pt1{_session.iter->first, 0}, pt2{next->first, 1};
+          // If either has value interpolate using default value.
```
hidmic, line 279: @arjo129 meta: ok with this behavior for a first pass, but it may not be appropriate for several use cases. Perhaps we can add a ticket to generalize this a bit (e.g. to allow constant extrapolation or simply return nullopt or whatever you can think of).

include/gz/math/TimeVaryingVolumetricGrid.hh
```diff
@@ -0,0 +1,152 @@
...
+  private: TimeVaryingVolumetricGridLookupField
+    <T, V, InMemorySession<T, P>> indices;
+
+  template<typename U, typename S, typename X>
+  friend class InMemoryTimeVaryingVolumetricGridFactory;
+};
+
+/// \brief Alias for Specialization of TimeVaryingVolumetricGrid which loads
+/// the whole of the dataset into memory.
+template<typename T, typename V, typename P>
```
hidmic, line 107: @arjo129 nit^2: it would be nice and convenient to use T as default for V and P. Then we can go from `InMemoryTimeVaryingVolumetricGrid<double, double, double>` to `InMemoryTimeVaryingVolumetricGrid<double>`. Same elsewhere applicable.

include/gz/math/TimeVaryingVolumetricGridLookupField.hh
```diff
@@ -0,0 +1,291 @@
...
+        {
+          // This happens we reach the end of time
+          return _session.iter->second.EstimateValueUsingTrilinear(
+            _position,
+            _values2,
+            _default);
+        }
+
+        /// Got nothing to interpolate. Out of bounds.
+        if (_interpolators[0].timeSlice.size() == 0
```
hidmic, line 239: @arjo129 FYI, safe to ignore: there's [`std::vector<T, Allocator>::empty`](https://en.cppreference.com/w/cpp/container/vector/empty) too.

src/TimeVaryingVolumetricGridLookupField_TEST.cc
```diff
@@ -0,0 +1,193 @@
...
+  TimeVaryingVolumetricGridLookupField<
+    double, double, InMemorySession<double, double>> timeVaryingField;
+
+  timeVaryingField.AddVolumetricGridField(0, scalarIndex);
+  timeVaryingField.AddVolumetricGridField(1, scalarIndex);
+
+  {
+    // Get data at T=0
+    auto session = timeVaryingField.CreateSession();
+    auto points = timeVaryingField.LookUp(session, Vector3d{0.5, 0.5, 0.5});
```
hidmic, line 47: @arjo129 nit^2: having a `const tolerance = Vector3d{0.5, 0.5, 0.5}` right above may improve readability a bit.

src/TimeVaryingVolumetricGridLookupField_TEST.cc
```diff
@@ -0,0 +1,193 @@
...
+    ASSERT_EQ(points.size(), 2);
+    ASSERT_EQ(points[0].time, 0);
+    ASSERT_EQ(points[1].time, 1);
+    auto res = timeVaryingField.EstimateQuadrilinear<double>(
```
hidmic, line 51: @arjo129 nit: I'd think it's not necessary to specify the template parameter here? Compiler should be able to infer X from arguments.

hidmic (approved):

hidmic:
@arjo129 we got some linter errors in CI

arjo129:
Seems like homebrew is not passing for unknown reasons.

hidmic:
> Seems like homebrew is not passing for unknown reasons.

@scpeters any ideas? We'd like to get this into `main` soon.

hidmic:
@arjo129 looks like it was a hiccup. We're good to go.

PR #455, opened 2022-07-06
Add Matrix6 class

chapulina:
# 🎉 New feature

* Part of https://github.com/gazebosim/gz-sim/issues/1462
* Alternative to
    * https://github.com/gazebosim/gz-math/pull/442
    * https://github.com/gazebosim/gz-math/pull/447
* Builds on top of https://github.com/gazebosim/gz-math/pull/453

## Summary

Add a new `Matrix6` class to be used to store added mass (see https://github.com/gazebosim/gz-sim/issues/1462).

The class is an extension of the existing `Matrix4` class, with these differences:

* It has 2 more rows and columns
* It has the improvements that have been made to `Matrix4` on `main`:
    * #206 and #376
    * #283
    * #293 
* It doesn't have any of the APIs that don't apply to a 6x6 matrix, such as conversion to pose, affine transforms, etc.
* It has the bare minimum functionality that's needed by added mass, plus some things that were trivial to implement. Notably, it's missing `Determinant` and `Inverse`
* It has a `Submatrix` function, as requested in https://github.com/gazebosim/gz-math/pull/442#issuecomment-1175427987
* I did some future-proofing to make it easier to port this to `main`, by using `GZ_` header guards, removing redundant `ignition::math::` namespaces, etc.

## Test it

Run the added tests.

## Checklist
- [x] Signed all commits for DCO
- [x] Added tests
- [ ] Added example and/or tutorial
- [x] Updated documentation (as needed)
- [x] Updated migration guide (as needed)
- [x] Consider updating Python bindings (if the library has them)
- [x] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [x] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸

PR #459, opened 2022-07-11
Add fluid added mass to inertial

chapulina:
# 🎉 New feature

## Summary

Added fluid added mass support to the inertial class, as described on the [Fluid Added Mass Proposal](http://sdformat.org/tutorials?tut=added_mass_proposal&branch=chapulina/added_mass).

I also had to add addition operators to `Matrix6`, because I forgot the one API I needed when I created the class in #455 🤦🏽 

## Test it

Take a look at the new tests.

## TODO

- [ ] The current implementation doesn't take into account the inertial pose's orientation. I think this may be a mistake. I'm going to revisit this as I open downstream PRs.
- [ ] Python bindings

## Checklist
- [x] Added tests


include/gz/math/Inertial.hh
```diff
@@ -123,6 +163,51 @@ namespace gz
...
+      /// \brief Spatial mass matrix for body B. It does not include fluid
+      /// added mass.
+      /// The matrix is arranged as follows:
+      ///
+      /// | m          0          0          0           m * Pz    -m * Py |
+      /// | 0          m          0          -m * Pz    0           m * Px |
+      /// | 0          0          m           m * Py    -m * Px    0       |
+      /// | 0          -m * Pz     m * Py    Ixx        Ixy        Ixz     |
+      /// |  m * Pz    0          -m * Px    Ixy        Iyy        Iyz     |
+      /// | -m * Py     m* Px     0          Ixz        Iyz        Izz     |
```
scpeters, line 175: as noted in https://github.com/gazebosim/sdf_tutorials/pull/76#discussion_r914144751, the spatial matrix convention used by [dartsim](https://github.com/dartsim/dart/blob/c5ed203bba13c427a4525dfd4b903080c2d34af6/dart/dynamics/Inertia.cpp#L409-L425) and [drake](https://drake.mit.edu/doxygen_cxx/group__multibody__spatial__inertia.html) (and based on [Featherstone's convention](https://homes.cs.washington.edu/~todorov/courses/amath533/FeatherstoneSlides.pdf)) uses `[p q r x y z]`, which differs from the `[x y z p q r]` convention used here. What is the plan for translating these matrices? It wouldn't be too hard using the `SubMatrix` API, but it's worth thinking about

include/gz/math/Inertial.hh
```diff
@@ -328,6 +413,9 @@ namespace gz
       /// \brief Pose offset of center of mass reference frame relative
       /// to a base frame.
       private: Pose3<T> pose;
+
+      /// \brief Fluid added mass.
+      private: Matrix6<T> addedMass;
```
scpeters, line 418: can this be a `std::optional<Matrix6<T>`?  it will be easier to detect if it has never been set

include/gz/math/Inertial.hh
chapulina: I was planning on rearranging the terms as needed when the value is passed to the physics engine. We can add helper functions here if you think that would help.

I'm not opposed to swapping things to follow the Featherstone convention, but putting translation above rotation is more in-line with Gazebo and ROS conventions, as well as most of the maritime literature I've looked at.

include/gz/math/Inertial.hh
chapulina: Good idea, done in [b22d60f](https://github.com/gazebosim/gz-math/pull/459/commits/b22d60f3106734fad452fd5fab6975edb34df8a7)

include/gz/math/Inertial.hh
scpeters: ok, I just wanted to make sure we've thought about it. I think the `SubMatrix` API should be enough

include/gz/math/Inertial.hh
```diff
@@ -112,6 +133,28 @@ namespace gz
...
+      public: bool SetFluidAddedMass(const Matrix6<T> &_m)
+      {
+        this->addedMass = _m;
+        return this->addedMass.value() == this->addedMass.value().Transposed();
+      }
+
+      /// \brief Get the fluid added mass matrix.
+      /// \return The added mass matrix. It will be nullpot of the added mass
+      /// was never set.
+      public: std::optional< Matrix6<T> > FluidAddedMass() const
```
scpeters, line 153: return a `const` reference?

include/gz/math/Inertial.hh
```diff
@@ -112,6 +133,28 @@ namespace gz
...
+      /// \return True if the Matrix6 is symmetric.
+      /// \sa SpatialMatrix
+      public: bool SetFluidAddedMass(const Matrix6<T> &_m)
+      {
+        this->addedMass = _m;
+        return this->addedMass.value() == this->addedMass.value().Transposed();
+      }
+
+      /// \brief Get the fluid added mass matrix.
+      /// \return The added mass matrix. It will be nullpot of the added mass
```
scpeters, line 151:
```suggestion
      /// \return The added mass matrix. It will be nullopt if the added mass
```

include/gz/math/Inertial.hh
chapulina: Done in [de61ec1](https://github.com/gazebosim/gz-math/pull/459/commits/de61ec19b697f2c9b97febb36cf0b460c3edd986). I haven't found any example code doing this though, but I think it should provide the functionality of `std::optional` while preventing a copy.

include/gz/math/Inertial.hh
```diff
@@ -112,6 +133,28 @@ namespace gz
...
+      /// \return True if the Matrix6 is symmetric.
+      /// \sa SpatialMatrix
+      public: bool SetFluidAddedMass(const Matrix6<T> &_m)
+      {
+        this->addedMass = _m;
+        return this->addedMass.value() == this->addedMass.value().Transposed();
+      }
+
+      /// \brief Get the fluid added mass matrix.
+      /// \return The added mass matrix. It will be nullopt of the added mass
```
scpeters, line 151:
```suggestion
      /// \return The added mass matrix. It will be nullopt if the added mass
```

include/gz/math/Inertial.hh
```diff
@@ -123,6 +166,52 @@ namespace gz
         return R * this->massMatrix.Moi() * R.Transposed();
       }
 
+      /// \brief Spatial mass matrix for body B. It does not include fluid
```
scpeters, line 169: this matrix should be expressed in frame F to match the `Moi()` API just above, and the documentation should explicitly mention that

include/gz/math/Inertial.hh
```diff
@@ -123,6 +166,52 @@ namespace gz
...
+      /// \return The body's 6x6 inertial matrix.
+      /// \sa SpatialMatrix
+      public: Matrix6<T> BodyMatrix() const
+      {
+        Matrix6<T> result;
+
+        result.SetSubmatrix(Matrix6<T>::TOP_LEFT,
+            Matrix3<T>::Identity * this->massMatrix.Mass());
+
+        result.SetSubmatrix(Matrix6<T>::BOTTOM_RIGHT, this->massMatrix.Moi());
```
scpeters, line 189: `this->massMatrix.Moi()` is expressed in frame `Bi`, but this API should return a matrix in frame `F`
```suggestion
        result.SetSubmatrix(Matrix6<T>::BOTTOM_RIGHT, this->Moi());
```

include/gz/math/Inertial.hh
```diff
@@ -123,6 +166,52 @@ namespace gz
...
+            0, -z, y,
+            z, 0, -x,
+            -y, x, 0) * this->massMatrix.Mass();
+        result.SetSubmatrix(Matrix6<T>::BOTTOM_LEFT, skew);
+        result.SetSubmatrix(Matrix6<T>::TOP_RIGHT, skew.Transposed());
+
+        return result;
+      }
+
+      /// \brief Spatial mass matrix, which includes the body's inertia, as well
```
scpeters, line 204: the documentation of this API should state in which frame the API is expressed, and that frame should be `F` to match the behavior of the `Moi()` function

the constructor currently says that the added mass matrix is expressed in frame `Bi`. would it be easier to require that the added mass matrix is expressed in Frame `F`?

include/gz/math/Inertial.hh
chapulina: Thanks. While implementing this I went back and forth between the frames, but I agree that the inertial class' API should consistently return values that take the pose into account. Updated in deac75cebb3a89576101407e79bb9aac853766b7.

include/gz/math/Inertial.hh
chapulina: deac75cebb3a89576101407e79bb9aac853766b7

include/gz/math/Inertial.hh
chapulina: Updated docs in deac75cebb3a89576101407e79bb9aac853766b7.

> would it be easier to require that the added mass matrix is expressed in Frame F?

It is definitely easier for me to implement, because I'm not sure how I'd transform a matrix from the CoM frame to the link frame. But it feels to me that if we do it like that, the API will be a bit less consistent, because the moments of inertia are expressed in the CoM frame (Bi).

I'll look around for a bit and see if I find a way of transforming the added mass matrix from one frame to another.

src/Inertial_TEST.cc
```diff
@@ -735,3 +739,124 @@ TEST(Inertiald_Test, SubtractionInvalid)
     EXPECT_TRUE((i2 - i1).MassMatrix().IsValid());
   }
 }
+
+TEST(Inertiald_Test, BodyMatrix)
+{
+  math::MassMatrix3d massMatrix(100, {1, 2, 3}, {4, 5, 6});
+  math::Pose3d com{7, 8, 9, 0, 0, 0};
```
scpeters, line 746: I would suggest a non-zero pose rotation so we can confirm how we expect that to work (I think it only affects the bottom right sub matrix)

include/gz/math/Inertial.hh
chapulina: Looking at other references, I'm thinking it's best to leave it in the link frame. I'm also not clear how I'd transform it.

scpeters: 👍

scpeters (approved):
I clarified one of the new test comments in https://github.com/gazebosim/gz-math/pull/459/commits/2ad032f2887fb048144ea7fca3e771e532ca31ce and from there it looks good to me!

PR #460, opened 2022-07-12
ign -> gz CMake, Python, Partial Source, and File Migrations : gz-math

methylDragon:
See: https://github.com/gazebo-tooling/release-tools/issues/765


configure.bat
```diff
@@ -4,7 +4,7 @@
 
 :: Install dependencies
 call %win_lib% :download_unzip_install eigen3-3.3.4.zip
-call %win_lib% :install_ign_project gz-cmake gz-cmake3
+call %win_lib% :install_gz_project gz-cmake gz-cmake3
```
scpeters, line 7: this file can be removed

scpeters:
I recommend using branches that start with `ci_matching_branch/` for pull requests that require branches from other pull requests (in this case https://github.com/gazebosim/gz-cmake/pull/275). I can advise on how to set this up so that macOS and windows CI will use the appropriate branches

PR #466, opened 2022-07-14
ign -> gz Tick-tock Python Libs : gz-math

methylDragon:
Part of overarching: https://github.com/gazebo-tooling/release-tools/issues/698
Follow-up to: https://github.com/gazebo-tooling/release-tools/issues/765

In https://github.com/gazebo-tooling/release-tools/issues/765, gz-math Python bindings were tick-tocked to allow use of `ignition` and `gz`. But the symlinks that are being created on the debbuilder are malformed, even though local builds work fine...

This PR fixes that by abusing relative links (it breaks on absolute links for some reason.)

Compare the built `.deb` from this
- **broken**: https://build.osrfoundation.org/job/ign-math7-debbuilder/741/
  - (the symlink points to one layer too shallow) (`../../../../usr/lib/XXX`, when it should be `../../../../../usr/lib/XXX`)
- **fixed**: https://build.osrfoundation.org/job/ign-math7-debbuilder/745/
  - (symlink now points to `../gz/XXX`)

Test downstream build using fixed debbuild: [Build Status](https://build.osrfoundation.org/job/sdformat-ci-pr_any-ubuntu_auto-amd64/3858/)
Build is unstable because of deprecation warnings, not test failures, which this fixes!

### DIFF
--- a/src/python_pybind11/CMakeLists.txt
+++ b/src/python_pybind11/CMakeLists.txt
@@ -110,19 +110,20 @@ function(configure_build_install_location _library_name)
   # Install Python library symlinks
   if(${GZ_PYTHON_INSTALL_PATH} MATCHES "gz$")
     cmake_policy(SET CMP0087 NEW)  # Allow evaluation of generator expressions in install(CODE )
+    file(MAKE_DIRECTORY "${PROJECT_BINARY_DIR}/ignition")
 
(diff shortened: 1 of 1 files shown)


methylDragon:
Build is unstable because of deprecation warnings, not test failures, which this fixes!

adityapande-1995 (approved):

PR #475, opened 2022-07-27
VolumetricGridLookupField: generalize index type, add constructor 

arjo129:
This PR generifies the index term. It also adds a constructor which is used by #456.

Signed-off-by: Arjo Chakravarty <arjo@openrobotics.org>


src/VolumetricGridLookupField_TEST.cc
```diff
@@ -48,6 +48,29 @@ TEST(VolumetricGridLookupField, CheckInterpolationExact)
...
+  std::vector<std::size_t> indices;
+  const double stride_x = 1, stride_y = 5, stride_z = 10;
+  for(double x = 0; x < 300; x += stride_x)
+  {
+    for(double y = 0; y < 300; y += stride_y)
+    {
+      for(double z = 0; z < 300; z += stride_z)
+      {
+        cloud.emplace_back(x, y, z);
+        indices.push_back(0);
```
hidmic, line 64: @arjo129 fyi: you can also do `indices.resize(cloud.size(), 0)` after the triple nested loop.

PR #479, opened 2022-08-09
ign -> gz : Remove redundant namespace references

methylDragon:
Part of: https://github.com/gazebo-tooling/release-tools/issues/784

### DIFF
--- a/src/PID.cc
+++ b/src/PID.cc
@@ -136,7 +136,7 @@ double PID::Update(const double _error,
                    const std::chrono::duration<double> &_dt)
 {
   if (_dt == std::chrono::duration<double>(0) ||
-      ignition::math::isnan(_error) || std::isinf(_error))
+      isnan(_error) || std::isinf(_error))
   {
     return 0.0;
   }
--- a/src/RotationSpline.cc
+++ b/src/RotationSpline.cc
@@ -72,9 +72,9 @@ Quaterniond RotationSpline::Interpolate(const unsigned int _fromIndex,
   }
 
   // Fast special cases
-  if (math::equal(_t, 0.0))
+  if (equal(_t, 0.0))
     return this->dataPtr->points[_fromIndex];
-  else if (math::equal(_t, 1.0))
+  else if (equal(_t, 1.0))
     return this->dataPtr->points[_fromIndex + 1];
 
   // double interpolation
--- a/src/AxisAlignedBox.cc
+++ b/src/AxisAlignedBox.cc
@@ -44,8 +44,8 @@ AxisAlignedBox::AxisAlignedBox(double _vec1X, double _vec1Y, double _vec1Z,
   this->dataPtr->min.Set(_vec1X, _vec1Y, _vec1Z);
   this->dataPtr->max.Set(_vec2X, _vec2Y, _vec2Z);
 
-  this->dataPtr->min.Min(math::Vector3d(_vec2X, _vec2Y, _vec2Z));
-  this->dataPtr->max.Max(math::Vector3d(_vec1X, _vec1Y, _vec1Z));
+  this->dataPtr->min.Min(Vector3d(_vec2X, _vec2Y, _vec2Z));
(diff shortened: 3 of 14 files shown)


src/PID.cc
```diff
@@ -136,7 +136,7 @@ double PID::Update(const double _error,
                    const std::chrono::duration<double> &_dt)
 {
   if (_dt == std::chrono::duration<double>(0) ||
-      ignition::math::isnan(_error) || std::isinf(_error))
+      isnan(_error) || std::isinf(_error))
```
chapulina, line 139: This one may be worth differentiating, so people reading the code know that it isn't `std::isnan`

src/RotationSpline.cc
```diff
@@ -72,9 +72,9 @@ Quaterniond RotationSpline::Interpolate(const unsigned int _fromIndex,
   }
 
   // Fast special cases
-  if (math::equal(_t, 0.0))
+  if (equal(_t, 0.0))
```
chapulina, line 75: Here too, this could be confused with `std::equal`, so it may be worth keeping for readability.

chapulina (approved):
LGTM, I just have minor comments about being more explicit when the functions could be confused with those in the `std::` namespace. It's not a big deal though, because we always use the `std` namespace explicitly and I believe the compiler (at least the ones we support) would complain if there was an ambiguity.

scpeters:
the results here seem inconsistent; sometimes the `math::` is removed and sometimes not

src/PID.cc
methylDragon:
```suggestion
      ignition::math::isnan(_error) || std::isinf(_error))
```

src/RotationSpline.cc
methylDragon:
```suggestion
  if (math::equal(_t, 0.0))
```

methylDragon:
> the results here seem inconsistent; sometimes the `math::` is removed and sometimes not

It's dependent on the appearance of a preceding `using namespace math;` directive (and a few escape hatches).
I couldn't get the namespaces removed for class declarations and function arguments and still have it compile as well :o

methylDragon:
> LGTM, I just have minor comments about being more explicit when the functions could be confused with those in the `std::` namespace. It's not a big deal though, because we always use the `std` namespace explicitly and I believe the compiler (at least the ones we support) would complain if there was an ambiguity.

Speaking of which, I have an explicit escape hatch to **not** remove the following namespaces :D
- `std`
- `google::protobuf`
- `details`

scpeters:
> > the results here seem inconsistent; sometimes the `math::` is removed and sometimes not
> 
> It's dependent on the appearance of a preceding `using namespace math;` directive (and a few escape hatches). I'm not sure if there's a way to programatically determine when it would be good to add in a `using` directive..
> 
> I couldn't get the namespaces removed for class declarations and function arguments and still have it compile as well :o

I think extra namespaces are sometimes included as a style choice or to make it more obvious where the given method is being called from, so this PR is kind of a big set of style changes. I would have no qualms about this if it only removed `ignition::` namespaces and left the `math::` namespaces in place.

That said, I'm willing to accept these changes but with hesitation

chapulina:
> sometimes included as a style choice or to make it more obvious where the given method is being called from

We haven't been enforcing a specific style when it comes to using namespaces. We've actually not been following the [Google style guide](https://google.github.io/styleguide/cppguide.html#Namespaces) too closely, because we use `using namespace ignition` in the `.cc` files instead of wrapping the file in `namespace ignition {}`. I didn't find any recommendations there about repeating a namespace within itself. So I don't think we need to block this PR on this.

We just need to be careful not to call top-level global functions by mistake as it happened in https://github.com/gazebosim/gz-common/pull/414#discussion_r941876670, because that actually changes the code.

scpeters:
> So I don't think we need to block this PR on this.

agreed; I just wanted to express a concern

PR #483, opened 2022-08-12
ign -> gz Migrate Ignition Headers : gz-math

methylDragon:
See: https://github.com/gazebo-tooling/release-tools/issues/784


scpeters:
I've been looking at the abichecker CI failures, and I think it's a problem that the classes are defined in the `gz` namespace. I've checked the symbols using `nm` on a `.so` file from this branch, and there are no `ignition::` symbols in the library, so I believe this approach breaks ABI.

We can use the `namespace gz { using namespace ignition; }` trick, but I think the class definitions need to remain in the `ignition` namespace

methylDragon:
> I've been looking at the abichecker CI failures, and I think it's a problem that the classes are defined in the `gz` namespace. I've checked the symbols using `nm` on a `.so` file from this branch, and there are no `ignition::` symbols in the library, so I believe this approach breaks ABI.
> 
> We can use the `namespace gz { using namespace ignition; }` trick, but I think the class definitions need to remain in the `ignition` namespace

I got this to work, and I'll copy paste my notes regarding my approach here:

- To not break ABI,  the headers need to be using the ignition namespace, and the ignition namespace needs to be the "primary" namespace. So I'm back to using reverse redirection (any gz namespaces will be using ignition )
  - To this end, I've put the namespace directives in config.hh for BOTH gz and ignition
- For msgs, since generation happens automatically with protoc, I can't force the headers to use ignition, so I can't migrate them to gz. msgs will hence not be migrated in Citadel (and Fortress)

- I've tried my best to migrate in-source namespace declarations to gz, but this cannot happen in a sweeping manner across the board, so there will be some interspersing of ignition and gz, but the diffs should be reduced, which still fits with the goal of making forward/backports easier. Some instances of these exceptions include:
  - Some packages don't use config.hh , so can't pick up the namespace redirections (common, plugin, fuel-tools)
  - Sometimes the redirections don't apply to namespace XXX declarations, so those aren't migrated
  - Some gz:: usages don't work (e.g. inside an explicitly ignition namespaced class)

scpeters:
nice work on fixing the ABI checker.

how do you feel about squashing the last commit (https://github.com/gazebosim/gz-math/pull/483/commits/0fc7ef239a7187de3e61f450842643f4c41e792b) into the second commit (https://github.com/gazebosim/gz-math/pull/483/commits/9309be044094233b34a5b035f24c2063b23ed6b0)?

methylDragon:
> nice work on fixing the ABI checker.
> 
> how do you feel about squashing the last commit ([0fc7ef2](https://github.com/gazebosim/gz-math/commit/0fc7ef239a7187de3e61f450842643f4c41e792b)) into the second commit ([9309be0](https://github.com/gazebosim/gz-math/commit/9309be044094233b34a5b035f24c2063b23ed6b0))?

Once everything is ok, I'll squash all the commits into the second commit (leaving the first commit (the move) alone) for the rebase-merge!

scpeters (approved):

test/integration/deprecated_TEST.cc
```diff
@@ -0,0 +1,27 @@
...
+*/
+
+#include <gtest/gtest.h>
+#include <ignition/math/Angle.hh>
+
+/////////////////////////////////////////////////
+// Make sure the ignition namespace still works
+TEST(Deprecated, IgnitionNamespace)
+{
+  gz::math::Angle angle;
```
chapulina, line 25: Should this be 

```suggestion
  ignition::math::Angle angle;
```

include/ignition/math/config.hh
```diff
@@ -0,0 +1,48 @@
...
+/* #undef IGNITION_MATH_BUILD_TYPE_RELEASE */
+
+namespace ignition
+{
+}
+
+namespace gz
+{
+  using namespace ignition;
+}
```
chapulina, line 46: Do we need to repeat this here? It's already including `gz/math/config.hh`

CMakeLists.txt
```diff
@@ -16,7 +16,10 @@ find_package(ignition-cmake2 2.13.0 REQUIRED)
 #============================================================================
 set (c++standard 17)
 set (CMAKE_CXX_STANDARD 17)
-ign_configure_project(VERSION_SUFFIX)
+ign_configure_project(
+  REPLACE_IGNITION_INCLUDE_PATH gz/math
```
chapulina, line 20: I'm not completely sure about the consequences of this change, could this break anyone downstream, @scpeters ?

chapulina (commented):
Great work, the general idea is ok to me, I just have some minor comments / questions

CMakeLists.txt
scpeters: this changes where the configured header files (`math.hh`, `math/config.hh`, `math/Export.hh`, etc) are installed. The redirection headers installed in `include/ignition` should keep working any existing code that includes `ignition/math*`. I think it's safe to do; the abichecker job is happy

include/ignition/math/config.hh
methylDragon: Ah, you're right. It's a leftover from when I wasn't including it on the `gz` one... :sweat: 
I need to remove them from the entire stack

PR #498, opened 2022-09-08
Bugfix/matrix buffer protocol

willstott101:
## Summary
```
>>> list(np.array(Matrix4d()))
  File "<stdin>", line 1, in <module>
TypeError: iteration over a 0-d array
```

Now:
```
>>> m = Matrix4d(0.0, 0.1, 0.2, 0.3, 10,11,12,13,20,21,22,23,30,31,32,33)
0 0.1 0.2 0.3 10 11 12 13 20 21 22 23 30 31 32 33
>>> memoryview(m)[0,0]
0.0
>>> memoryview(m)[2,1]
21.0
>>> m(2,1)
21.0
>>> np.array(m)[2,1]
21.0
>>> np.array(m).tolist()
[[0.0, 0.1, 0.2, 0.3], [10.0, 11.0, 12.0, 13.0], [20.0, 21.0, 22.0, 23.0], [30.0, 31.0, 32.0, 33.0]]
```
## Checklist
- [x] Added tests

I haven't signed off yet cause I want feedback on making `data` non-private - is there another way to access that data from the python header - some sort of `friend` thing?

### DIFF
--- a/include/gz/math/Matrix3.hh
+++ b/include/gz/math/Matrix3.hh
@@ -636,7 +636,7 @@ namespace gz
       }
 
       /// \brief the 3x3 matrix
-      private: T data[3][3];
+      T data[3][3];
     };
 
     namespace detail {
--- a/include/gz/math/Matrix4.hh
+++ b/include/gz/math/Matrix4.hh
@@ -844,7 +844,7 @@ namespace gz
       }
 
       /// \brief The 4x4 matrix
-      private: T data[4][4];
+      T data[4][4];
     };
 
     namespace detail {
--- a/include/gz/math/Matrix6.hh
+++ b/include/gz/math/Matrix6.hh
@@ -553,7 +553,7 @@ namespace gz
       }
 
       /// \brief The 6x6 matrix
-      private: T data[MatrixSize][MatrixSize];
+      T data[MatrixSize][MatrixSize];
     };
 
     namespace detail {
--- a/src/python_pybind11/src/Matrix3.hh
+++ b/src/python_pybind11/src/Matrix3.hh
@@ -123,7 +123,18 @@ void helpDefineMathMatrix3(py::module &m, const std::string &typestr)
     .def_readonly_static("IDENTITY", &Class::Identity, "Identity matrix")
     .def_readonly_static("ZERO", &Class::Zero, "Zero matrix")
     .def("__str__", toString)
-    .def("__repr__", toString);
+    .def("__repr__", toString)
+    .def_buffer([](Class &self) -> py::buffer_info {
+      return py::buffer_info(
+          self.data,  // Pointer to buffer
+          sizeof(T),  // Size of one scalar
+          py::format_descriptor<T>::format(),
+          2,  // Number of dimensions
+          { 3, 3 },  // Buffer dimensions
+          // Strides (in bytes) for each dimension
+          { sizeof(T) * 3, sizeof(T) }
+      );
+    });
 }
 }  // namespace python
 }  // namespace math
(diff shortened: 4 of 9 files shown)


willstott101:
Hopefully failing tests here: https://github.com/willstott101/gz-math/actions/runs/3016276863

adityapande-1995 (commented):
Hi ! Could you please add signoffs to your commits ? https://gazebosim.org/docs/all/contributing#contributing-code

willstott101:
I mentioned in the original PR text:

> I haven't signed off yet cause I want feedback on making data non-private - is there another way to access that data from the python header - some sort of friend thing?

I'm not happy to sign this off yet so I won't sign off the commits.

willstott101:
Takes a really long time for the CI to roll through, here's latest runs (since they're not allowed to run here by the looks of it?)

Expected failing new tests: https://github.com/willstott101/gz-math/actions/runs/3016796871
After patches to fix those tests: https://github.com/willstott101/gz-math/actions/runs/3016795365

mjcarroll:
> since they're not allowed to run here by the looks of it?

Since it was your first time contributing they had to be manually approved, they will run now (and I don't believe you need approval in the future)

willstott101:
Ok finally happy (with everything but making `data` public).

* failing test: https://github.com/willstott101/gz-math/actions/runs/3021342616

Can now do this:
Resulting buffers are contiguous, index in the same way whether using numpy, memoryview, or `Matrix*.__call__`.

This means we can go from Matrix -> flat list with () requires contiguous to cast:
`memoryview(m).cast('b', (4 * 4 * 8,)).cast('d', (4 * 4,)).tolist()`

And matrix to structured list with (does not require contiguous):
`memoryview(m).tolist()`

mjcarroll:
> Ok finally happy (with everything but making data public).

Did you have a workaround for this yet?

willstott101:
Nope, I'm not much of a c++ guy, doubt I'll get to digging to be honest - but happy to update if someone has an idea

willstott101:
I've had a bit of a dig into this, and I don't see any way to make the python binding have access to the private member. It just doesn't seem possible with the way the templates are used.

I don't think making data public is all that bad, would you be tempted to accept this MR with making the data var public?

willstott101:
I've signed off the commits

azeey:
How about using accessor functions?

willstott101:
To what end? This PR is about exposing the underlying multi-dimensional array to python. Accessor functions won't enable that.

The buffer protocol is commonly used for converting to/from numpy arrays for instance. I happen to be using it to get the data as a python list, though that could of course be done slowly with accessor functions, exposing the underlying data using the buffer protocol is more flexible, potentially faster in some scenarios and is very common for matrix/array programming in python.

scpeters:
> I've had a bit of a dig into this, and I don't see any way to make the python binding have access to the private member. It just doesn't seem possible with the way the templates are used.
> 
> I don't think making data public is all that bad, would you be tempted to accept this MR with making the data var public?

I would prefer adding an accessor method that returns the `data` pointer, since it is easier to deprecate methods than member variables. I think we've avoided exposing the raw data pointer to help protect from invalid memory access, but it sounds like there's some demand for it, so we could consider add some warnings in the documentation to be careful with the memory access.

willstott101:
I see, I hadn't considered deprecation processes. That makes sense, thank you for explaining.

I understand there might be some hesitance with exposing the underlying array. Definitely rarely need to, particularly when used in a language with very fast (or even inlined) function calls. Sadly python isn't that :(

Would the return type simply be `T*`? Any appetite for `std::array<std::array<T, 3>, 3>&` instead? Either using `reinterpret_cast`, or actually changing the underlying member type.

Perhaps that would help reduce the likelyhood of using it incorrectly. And also make it easier to support a copying getter/setter pair if the reference returning method had to be deprecated.

willstott101:
Closing in favour of https://github.com/gazebosim/gz-math/pull/524

PR #503, opened 2022-09-14
Rename the python library as gz.math7

j-rivero:
# 🦟 Bug fix

## Summary
Allow side-by-side installations of python bindings by renaming the library to include the version in the name. From `ignition.math` or `gz.math` to `gz.math7`.

Particularly, the current approach is causing problems since it conflicts with `gz-math6` bindings since both install in `ignition/math`  path in the filesystem. This is blocking the installation of Fortress + Garden side-by-side.

 Checklist
- [x] Added tests

### DIFF
--- a/tutorials/cppgetstarted.md
+++ b/tutorials/cppgetstarted.md
@@ -7,7 +7,7 @@ Previous Tutorial: \ref install
 This tutorial describes how to get started using Gazebo Math with C++.
 
 We will run through an example that determines the distance between two
-points in 3D space. Start by creating a bare-bones main file using the
+points in 3D space. Start by creating a bare-bones `main.cpp` file using the
 editor of your choice.
 
 ```{.cpp}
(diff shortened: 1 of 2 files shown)


scpeters:
we'll need to make a prerelease after this is merged and update downstream python bindings in sdformat

Migration.md
```diff
@@ -19,6 +19,8 @@ release will remove the deprecated code.
       * CMake `-config` files
       * Paths that depend on the project name
 
+  1. Python library `ignition.math` namespaces should be replaced with `gz.math7`.
```
azeey, line 22:
```suggestion
  1. Python library imports such `import ignition.math` and `from ignition import math` should be replaced with `import gz.math7` and `from gz import math7`. Note the change from `ignition` to `gz` and the addition of the major version number as a suffix to the package name.
```

azeey (approved):

Migration.md
scpeters: should we recommend `from gz import math7 as math`?

j-rivero: > should we recommend from gz import math7 as math?

ouch, sorry I missed this one. I'm fine about using that recommendation, we can modify the Changelog but specially use the documentation and examples to do it.

scpeters: I didn't suggest it very quickly, but yes we can consider it for documentation

peci1:
It's definitely good to have the option to select the particular version. However, what is the maintenance plan for this? If there is no way for downstream code to use some convenience like `import gz.math`, keeping 3rd-party (and even 1st-party) code would become a nightmare...

PR #540, opened 2023-07-27
7 to main

ahcorde:
# ➡️ Forward port

Port 7 to main

Branch comparison: https://github.com/gazebosim/gz-math/compare/main...gz-math7


azeey:
This is currently targetting `gz-math7`. Should it target `main`?

ahcorde:
> This is currently targetting `gz-math7`. Should it target `main`?

Done

adityapande-1995 (commented):
Could you add small tests for the newly added functions to make codecov happy ?
Not sure why ABI checker is failing here.

azeey:
> Could you add small tests for the newly added functions to make codecov happy ? Not sure why ABI checker is failing here.

The ABI checker is failing because this PR was originally targeting `gz-math7`. Now that it's targetting `main`. we can ignore the ABI checker (it's okay to break ABI for `main`).
Also, since this is a forward port of changes in gz-math7, we generally don't add new tests. It should be sufficient to check that the changes from the older branch are compatible with the new branch. If there's a particular concern there, maybe a test would be warranted.

mjcarroll:
I think this one is good to go.  ABI checker is a non-issue and we can increase coverage in a later PR, as this is just a forward port.

mjcarroll (approved):

adityapande-1995 (approved):

Issue #548, opened 2023-08-22
Add MecanumDriveOdometry Python wrapper


ahcorde:
Related with this PR https://github.com/gazebosim/gz-math/pull/486

ahcorde:
Related PR https://github.com/gazebosim/gz-math/pull/549

PR #549, opened 2023-08-23
Added MecanumDriveOdometry Python wrapper

ahcorde:
# 🎉 New feature

Closes https://github.com/gazebosim/gz-math/issues/548

## Summary

Added MecanumDriveOdometry

## Checklist
- [x] Added tests
- [x] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers


azeey:
Is it okay if we wait till after Harmonic for this?

ahcorde:
this is targeting Fortress, but we can wait no problem

azeey:
> this is targeting Fortress, but we can wait no problem

I understand, but I think we should focus our review efforts on `beta` labeled PRs.

ahcorde:
friendly ping @azeey @scpeters or @adityapande-1995

src/python_pybind11/src/MecanumDriveOdometry.cc
```diff
@@ -0,0 +1,75 @@
...
+{
+namespace python
+{
+void defineMathMecanumDriveOdometry(py::module &m, const std::string &typestr)
+{
+  using Class = gz::math::MecanumDriveOdometry;
+  std::string pyclass_name = typestr;
+  py::class_<Class>(m,
+                    pyclass_name.c_str(),
+                    py::buffer_protocol(),
```
azeey, line 36: Do we need `py::buffer_protocol` here? If so, doesn't it require implementing `def_buffer` (https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html)

src/python_pybind11/src/MecanumDriveOdometry.cc
```diff
@@ -0,0 +1,75 @@
...
+                    py::dynamic_attr())
```
azeey, line 37: Why do we need `py::dynamic_attr`?

src/python_pybind11/test/MecanumDriveOdometry_TEST.py
```diff
@@ -0,0 +1,121 @@
...
+        wheelRadius = 0.5
+        wheelCircumference = 2 * math.pi * wheelRadius
+
+        # This is the linear distance traveled per degree of wheel rotation.
+        distPerDegree = wheelCircumference / 360.0
+
+        # Setup the wheel parameters, and initialize
+        odom.set_wheel_params(wheelSeparation, wheelRadius, wheelRadius,wheelRadius)
+        startTime = datetime.datetime.now()
+        odom.init(datetime.timedelta())
```
azeey, line 42: Should this use `startTime`?

src/python_pybind11/test/MecanumDriveOdometry_TEST.py
```diff
@@ -0,0 +1,121 @@
...
+
+        # Sleep for a little while, then update the odometry with the new wheel
+        # position.
+        time1 = startTime + datetime.timedelta(milliseconds=100)
+        odom.update(Angle(1.0 * math.pi / 180),
+                    Angle(1.0 * math.pi / 180),
+                    Angle(1.0 * math.pi / 180),
+                    Angle(1.0 * math.pi / 180),
+                    time1 - startTime)
```
azeey, line 51: The C++ version uses just `time1` here.

src/python_pybind11/test/MecanumDriveOdometry_TEST.py
```diff
@@ -0,0 +1,121 @@
...
+        # straight line.
+        self.assertAlmostEqual(0.0, odom.angular_velocity().radian(), delta=1e-3)
+
+        # Sleep again, then update the odometry with the new wheel position.
+        time2 = time1 + datetime.timedelta(milliseconds=100)
+        odom.update(Angle(2.0 * math.pi / 180),
+                    Angle(2.0 * math.pi / 180),
+                    Angle(2.0 * math.pi / 180),
+                    Angle(2.0 * math.pi / 180),
+                    time2 - startTime)
```
azeey, line 67: This is also different from the C++ test

src/python_pybind11/test/MecanumDriveOdometry_TEST.py
```diff
@@ -0,0 +1,121 @@
...
+
+        # Setup the wheel parameters, and initialize
+        odom.set_wheel_params(wheelSeparation, wheelRadius, wheelRadius,wheelRadius)
+        startTime = datetime.datetime.now()
+        odom.init(datetime.timedelta())
+
+        # Sleep for a little while, then update the odometry with the new wheel
+        # position.
+        time1 = startTime + datetime.timedelta(milliseconds=100)
+        odom.update(Angle(1.0 * math.pi / 180),
```
azeey, line 47: Use `math.radians`

azeey (changes requested):

azeey (changes requested):
Looks good overall. Just a few issues with pybind11 options and differences with the C++ test

azeey (changes requested):
Looks good overall. Just a few issues with pybind11 options and differences with the C++ test

ahcorde:
@azeey Time is a little bit different in Python, you can take a look the `DiffDriveOdometry_TEST.py` test,

ahcorde:
friendly ping @azeey

azeey:
> @azeey Time is a little bit different in Python, you can take a look the `DiffDriveOdometry_TEST.py` test,

So from the [pybind11 docs](https://pybind11.readthedocs.io/en/stable/advanced/cast/chrono.html#provided-conversions), it looks like `std::chrono::steady_clock::time_point` is converted to `datetime.timedelta`. This makes it really inconvenient to use since `datetime.timedelta()` always returns a "0" time delta as opposed to the current monotonic (in C++ `steady_clock`) time. One option is to use `datetime.timedelta(time.monotonic())` to initialize `startTime` instead of `datetime.datetime.now()`.

src/python_pybind11/src/MecanumDriveOdometry.cc
azeey: I don't think this has been addressed

j-rivero:
@osrf-jenkins run tests

ahcorde:
@azeey I added your feedback

azeey:
> @azeey I added your feedback

Thanks. I think now the test can be updated to match the C++ version. Specifically, `odom.init(startTime)` instead of `odom.init(datetime.timedelta())` and remove `startTime` from `odom.update` calls.

ahcorde:
> Thanks. I think now the test can be updated to match the C++ version. Specifically, `odom.init(startTime)` instead of `odom.init(datetime.timedelta())` and remove `startTime` from `odom.update` calls.

@azeey Included feedback

src/python_pybind11/test/DiffDriveOdometry_TEST.py
```diff
@@ -38,7 +39,7 @@ def test_diff_drive_odometry(self):
 
         # Setup the wheel parameters, and initialize
         odom.set_wheel_params(wheelSeparation, wheelRadius, wheelRadius)
-        startTime = datetime.datetime.now()
+        startTime = datetime.timedelta(time.monotonic())
         odom.init(datetime.timedelta())
```
azeey, line 43:
```suggestion
        startTime = datetime.timedelta(time.monotonic())
        odom.init(startTime)
```

then update
`odom.update(..., ..., time1)`

src/python_pybind11/test/DiffDriveOdometry_TEST.py
```diff
@@ -71,7 +72,7 @@ def test_diff_drive_odometry(self):
         self.assertAlmostEqual(0.0, odom.angular_velocity().radian(), delta=1e-3)
 
         # Initialize again, and odom values should be reset.
-        startTime = datetime.datetime.now()
+        startTime = datetime.timedelta(time.monotonic())
         odom.init(datetime.timedelta())
```
azeey, line 76:
```suggestion
        startTime = datetime.timedelta(time.monotonic())
        odom.init(startTime)
```


then update
`odom.update(..., ..., time1)`

src/python_pybind11/test/DiffDriveOdometry_TEST.py
azeey, line 1: We need to make the `odom.init` and `odom.update` changes to `time2` as well.

azeey (approved):

PR #562, opened 2023-10-06
Bumps in ionic: use gz-cmake4, gz-utils3

azeey:
See https://github.com/gazebo-tooling/release-tools/issues/1027

### DIFF
--- a/src/Quaternion_TEST.cc
+++ b/src/Quaternion_TEST.cc
@@ -326,11 +326,11 @@ TEST(QuaternionTest, MathAxis)
   math::Quaterniond q(GZ_PI*0.1, GZ_PI*0.5, GZ_PI);
 
 GZ_UTILS_WARN_IGNORE__DEPRECATED_DECLARATION
-  // Deprecated in gz-math8
+  // Deprecated in gz-math7
   q.Axis(0, 1, 0, GZ_PI);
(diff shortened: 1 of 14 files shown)


src/Quaternion_TEST.cc
```diff
@@ -326,11 +326,11 @@ TEST(QuaternionTest, MathAxis)
   math::Quaterniond q(GZ_PI*0.1, GZ_PI*0.5, GZ_PI);
 
 GZ_UTILS_WARN_IGNORE__DEPRECATED_DECLARATION
-  // Deprecated in gz-math7
+  // Deprecated in gz-math8
   q.Axis(0, 1, 0, GZ_PI);
   EXPECT_EQ(q, math::Quaterniond(6.12303e-17, 0, 1, 0));
 
-  // Deprecated in gz-math7
+  // Deprecated in gz-math8
```
scpeters, line 333: revert this comment change, and instead prepare to deprecate the related code

scpeters:
I merged the gz-utils3 pull requests and restarted the homebrew build since I think it may pass

scpeters:
> I merged the gz-utils3 pull requests and restarted the homebrew build since I think it may pass

homebrew build is happy; I think we need nightlies of gz-utils3 next

scpeters (approved):

PR #565, opened 2023-10-31
Fix visiblity for linux: move explicit instanciations to the header file

j-rivero:
# 🦟 Bug fix

## Summary

GCC was not happy with the use of a header only types in the explicit instantiations in order to generate the symbols in the library.  Fixed by moving them to the header.

Part of the changes for https://github.com/gazebosim/gz-cmake/pull/392
## Checklist
- [ ] Added tests

### DIFF
--- a/include/gz/math/MovingWindowFilter.hh
+++ b/include/gz/math/MovingWindowFilter.hh
@@ -92,7 +92,7 @@ namespace gz
   }
 
 
-/// The template explicit instanciations that does not use fundamental types
+/// The template explicit instantiations that does not use fundamental types
 /// (like Vector*) needs to be placed in this header file.
 template class GZ_MATH_VISIBLE MovingWindowFilter<int>;
 template class GZ_MATH_VISIBLE MovingWindowFilter<float>;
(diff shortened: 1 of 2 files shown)


j-rivero:
oh wonderful, there is now a different error in Mac and Windows.

include/gz/math/MovingWindowFilter.hh
```diff
@@ -90,6 +90,17 @@ namespace gz
     using MovingWindowFilterVector3f = MovingWindowFilter<Vector3f>;
     using MovingWindowFilterVector3d = MovingWindowFilter<Vector3d>;
   }
+
+
+/// The template explicit instantiations that does not use fundamental types
+/// (like Vector*) needs to be placed in this header file.
+template class GZ_MATH_VISIBLE MovingWindowFilter<int>;
```
azeey, line 97: I'm not sure if putting explicit instantiations in a header is the right thing to do (see https://stackoverflow.com/questions/5864401/does-explicit-template-instantiation-go-in-cpp-or-header-file). If every user of this header instantiates these classes, we'd have ODR issues. Also, we'd lose the benefit of explicit instantiation since every translation unit will be instantiating the templates.

azeey (changes requested):

j-rivero:
Current approach won't fix the issue and it triggers build warnings. I close this PR and open a new one.

PR #566, opened 2023-11-03
Use HIDE_SYMBOLS_BY_DEFAULT, fix MovingWindowFilter

j-rivero:
# 🦟 Bug fix

As part of the works of defining `hidden by default` in https://github.com/gazebosim/gz-cmake/pull/392, this PR moves the whole MovingWindowFilter class to be header only. Reason for this is that GCC on Jammy seems buggy to me (clang11 on Jammy worked fine, also clang on Mac and MSVC on Windows deal with it just fine) when it needs to generate the symbols needed for the specialization of the class done for `<gz::math::Vector3i>`.  See efforts in #565.

## Checklist
- [ ] Added tests


azeey:
Since gz-cmake already has `HIDDEN_SYMBOLS_BY_DEFAULT`, maybe we should use that in all the libraries first, make sure all CI passes and then merge https://github.com/gazebosim/gz-cmake/pull/392. Thoughts?

scpeters:
> Since gz-cmake already has `HIDDEN_SYMBOLS_BY_DEFAULT`, maybe we should use that in all the libraries first, make sure all CI passes and then merge [gazebosim/gz-cmake#392](https://github.com/gazebosim/gz-cmake/pull/392). Thoughts?

this sounds like you are suggesting the following:

1. Pass `HIDE_SYMBOLS_BY_DEFAULT` to the `gz_configure_build` calls in every `gz-*` package
2. Test and verify that CI / nightly builds are happy
3. Remove `HIDE_SYMBOLS_BY_DEFAULT` from the `gz_configure_build` calls in every `gz-*` package
4. Merge https://github.com/gazebosim/gz-cmake/pull/392

If we merge https://github.com/gazebosim/gz-cmake/pull/392 before removing the `HIDE_SYMBOLS_BY_DEFAULT` parameters, I think it may introduce cmake deprecation warnings. I guess if we split out the deprecation of `HIDE_SYMBOLS_BY_DEFAULT`, then we could do things in a slightly different order, but it would require an extra gz-cmake pull request

scpeters:
> > Since gz-cmake already has `HIDDEN_SYMBOLS_BY_DEFAULT`, maybe we should use that in all the libraries first, make sure all CI passes and then merge [gazebosim/gz-cmake#392](https://github.com/gazebosim/gz-cmake/pull/392). Thoughts?
> 
> this sounds like you are suggesting the following:
> 
> 1. Pass `HIDE_SYMBOLS_BY_DEFAULT` to the `gz_configure_build` calls in every `gz-*` package
> 2. Test and verify that CI / nightly builds are happy
> 3. Remove `HIDE_SYMBOLS_BY_DEFAULT` from the `gz_configure_build` calls in every `gz-*` package
> 4. Merge [Use visibility hidden by default gz-cmake#392](https://github.com/gazebosim/gz-cmake/pull/392)
> 
> If we merge [gazebosim/gz-cmake#392](https://github.com/gazebosim/gz-cmake/pull/392) before removing the `HIDE_SYMBOLS_BY_DEFAULT` parameters, I think it may introduce cmake deprecation warnings. I guess if we split out the deprecation of `HIDE_SYMBOLS_BY_DEFAULT`, then we could do things in a slightly different order, but it would require an extra gz-cmake pull request

it sounds a little like overkill, but I think it would be the most thorough approach since it would run CI on every single package

azeey:
> it sounds a little like overkill, but I think it would be the most thorough approach since it would run CI on every single package

Maybe a little overkill, but my gut feeling is that it won't be straightforward to get all the packages to build and test properly with symbols hidden by default, so I'd like CI to run on the PRs rather than just local testing. If we can do that without using the `HIDE_SYMBOLS_BY_DEFAULT` approach, I'm all for it.

scpeters:
> > it sounds a little like overkill, but I think it would be the most thorough approach since it would run CI on every single package
> 
> Maybe a little overkill, but my gut feeling is that it won't be straightforward to get all the packages to build and test properly with symbols hidden by default, so I'd like CI to run on the PRs rather than just local testing. If we can do that without using the `HIDE_SYMBOLS_BY_DEFAULT` approach, I'm all for it.

I opened https://github.com/gazebosim/gz-utils/pull/115 for example. I think it will take some additional development from our infrastructure team to support testing all packages from source on Ubuntu in CI

include/gz/math/MovingWindowFilter.hh
```diff
@@ -90,6 +90,127 @@ namespace gz
...
+    return T(value);
+  }
+  else
+  {
+    auto value = this->sum / static_cast<double>(this->samples);
+    return T(value);
+  }
+}
+
+template class MovingWindowFilter<int>;
```
azeey, line 207: Do we still need these if it's a header-only?

scpeters:
I opened #567 to set `HIDE_SYMBOLS_BY_DEFAULT` confirmed that CI fails, then I merged the changes into this PR

include/gz/math/MovingWindowFilter.hh
j-rivero: You are right, I don't think so. [56f145f](https://github.com/gazebosim/gz-math/pull/566/commits/56f145f1cd10ad34d5403c5f51c720057343f96c)

azeey:
@j-rivero Looks like there are two homebrew jobs? And both of them are failing trying to install either `gz-math` or `ign-math8`, neither of which are available.

image

scpeters:
> @j-rivero Looks like there are two homebrew jobs? And both of them are failing trying to install either `gz-math` or `ign-math8`, neither of which are available.
> 
> image

I think it's due to https://github.com/gazebo-tooling/release-tools/pull/1054

scpeters:
> @j-rivero Looks like there are two homebrew jobs? And both of them are failing trying to install either `gz-math` or `ign-math8`, neither of which are available.
> 
> image

I pushed a whitespace change (https://github.com/gazebosim/gz-math/pull/566/commits/17970d787fc5e348c11c8da6a86a0ab2088f0ec1) in order to re-run CI, and it's happy now

scpeters:
I'll wait for @azeey to approve

azeey (approved):

PR #579, opened 2024-02-21
Color [] changed to call by reference

Avisheet:
## Summary
The issue you're describing is related to the operator[] overload for the Color class. The problem is that the operator is declared to return a float instead of a reference to a float. As a result, when you use the operator to access a component of the color and attempt to mutate it, the mutation doesn't apply to the actual color instance.

To fix this issue, you should modify the declaration of the operator[] to return a reference to a float. Here's the corrected declaration:
```
public: float& operator[](const unsigned int _index);
```
This modification allows you to use the operator to both access and mutate the individual components of the Color class.

## Checklist
- [ ] Added tests
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers (optional)


azeey:
Hello @Avisheet! Thanks for the contribution.  The change you made in the header looks correct, but it requires a corresponding change in `Color.cc`. Would you be able to make that change? Also, could you add a unit test?

scpeters:
we would probably need to retarget to `main` as well since this breaks API, unless we add an operator overload instead of changing the current one

Avisheet:
Hello @azeey .
Thank you for your feedback. I've made the necessary changes in Color.cc to align with the modification in Color.hh. Additionally, I have added a corresponding unit test to ensure the functionality is working as expected.

Please review the changes and let me know if everything looks good to you. If you have any further suggestions or if there's anything else you'd like me to address, please feel free to let me know.

Avisheet:
hello @scpeters , Can you elaborate what you are telling as i am unable to understand it properly

Avisheet:
@azeey I have done some more changes in the Vector3_TEST.cc file to ensure that we get an expected result as "Operators can be used to [strictly order] vectors."Do check them once and tell me the modifications I need to do if any.

azeey:
@Avisheet I've added a checklist to your PR description; it should have been prepopulated when you created the PR unless  you removed it. Please make sure to complete all relevant items on the checklist.

Avisheet:
Hey @azeey . The existing behavior is that we return NAN on unknown indices. But if we want to support assignment then it cannot return any such const rvalue . so, we probably have to break that existing functionality . How should we handle that ?

.vscode/c_cpp_properties.json
azeey, line 1: Please remove this file.

.vscode/settings.json
azeey, line 1: Please remove this file.

include/gz/math/Color.hh
```diff
@@ -159,14 +159,14 @@ namespace gz
...
-      public: float operator[](const unsigned int _index);
+      public: float& operator[](const unsigned int _index);
 
       /// \brief Array index operator, const version
       /// \param[in] _index Color component index(0=red, 1=green, 2=blue,
       /// 3=alpha)
       /// \return r, g, b, or a when _index is 0, 1, 2 or 3. A NAN_F value is
       /// returned if the _index is invalid
-      public: float operator[](const unsigned int _index) const;
+      public: const float& operator[](const unsigned int _index) const;
```
azeey, line 169: It's better to return a POD type by value, so please revert this change.

src/Color.cc
```diff
@@ -187,13 +188,28 @@ void Color::SetFromYUV(const float _y, const float _u, const float _v)
...
+    case 2:
+      return this->b;
+    case 3:
+      return this->a;
+    default:
+      break;
+  }
+
+  std::cerr << "Trying to read index " << _index << " of Color" << std::endl;
+  throw std::runtime_error("Index Error: Color index out of range");
```
azeey, line 208: We don't typically use exceptions in gz-math, so please remove this. But I do see the problem with what we should do if the _index is out of range since we can return the `NAN_F` by reference.  The best thing I can think of is to create a static float here and return that by value.

```suggestion
  static float invalidValue = NAN_F;
  return invalidValue;
```

src/Color.cc
```diff
@@ -187,13 +188,28 @@ void Color::SetFromYUV(const float _y, const float _u, const float _v)
...
 }
 
 //////////////////////////////////////////////////
-float Color::operator[](const unsigned int _index) const
+const float& Color::operator[](const unsigned int _index) const
```
azeey, line 212: Please revert this per my comment in the header.

src/python_pybind11/test/Color_TEST.py
```diff
@@ -176,7 +176,7 @@ def test_color(self):
         self.assertAlmostEqual(0.5, clr[1])
         self.assertAlmostEqual(1.0, clr[2])
         self.assertAlmostEqual(1.0, clr[3])
-        self.assertTrue(math.isnan(clr[4]))
+        # self.assertTrue(math.isnan(clr[4]))
```
azeey, line 179: revert

src/Color_TEST.cc
```diff
@@ -433,3 +433,32 @@ TEST(Color, HSV)
...
+        {
+            clr[4] = 0.1f;
+        }
+        catch( const std::runtime_error& e )
+        {
+            // and this tests that it has the correct message
+            EXPECT_STREQ( "Index Error: Color index out of range", e.what() );
+            throw;
+        }
+    }, std::runtime_error);
```
azeey, line 463: I think it would be sufficient to check that `clr[4]` is a NAN here.

azeey (changes requested):

arjo129:
Hi @Avisheet, was wondering what the status on this was. Do we intend shipping this change with ionic (code freeze 26/8)? If we can't come to a consensus by then I will remove the beta label.

Please address @azeey's feedback on PODs.

src/Color.cc
```diff
@@ -16,6 +16,7 @@
  */
 #include <cmath>
 #include <algorithm>
+#include<iostream>
```
adityapande-1995, line 19: Do we need to include this ?

arjo129:
Hi @Avisheet as we haven't heard back from you, I'm removing the beta label. Both @adityapande-1995 and @azeey 's remarks need addressing.

arjo129:
Hey @Avisheet you don't have to close the PR if you'd like we can merge this in after the code freeze if you can't address it immediately.

Avisheet:
Hey @arjo129 , sorry for the inconvenience, but as I have some of my major projects going on, I won't be able to contribute to it for 1 more month.

Avisheet:
@arjo129 Yep, it's fine to release it with another gazebo release. I will be working on it for sure, but it's just I was a bit stuck with my major project.

Avisheet:
Hey, I had a doubt that how can I check and run all the test cases on my own device, before giving the PR, so that I can know where the error is, after doing some changes in code.

azeey:
I will go ahead and close this PR since there has been no progress in a long time.

PR #580, opened 2024-02-22
Expose non-const reference to edges in Graph.hh

### LINKED ISSUE #99: Graph.hh doesn't expose non-const reference to Edges
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-math/issues/99)) by Steve Peters (Bitbucket: [Steven Peters](https://bitbucket.org/%7B2ccfed09-18b8-4921-8d58-15ef01092802%7D/), GitHub: [scpeters](https://github.com/scpeters)).**

# Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Checked the Q&A board for common solutions: http://answers.gazebosim.org
    * Checked that your issue isn't already filed.
    * Checked that there is not already an Ignition package that provides the described functionality: https://ignitionrobotics.org/libs

# Description 

The [Graph::VertexFromId function](https://github.com/ignitionrobotics/ign-math/blob/5296217a8625994d07534f31551a899cf60a7a94/include/ignition/math/graph/Graph.hh#L609:622) has two versions that return a `const` and non-`const` reference to a Vertex. However there is only a `const` version of the [Graph::EdgeFromId function](https://github.com/ignitionrobotics/ign-math/blob/5296217a8625994d07534f31551a899cf60a7a94/include/ignition/math/graph/Graph.hh#L674), which makes it hard to modify Edge Data().

Likewise, the [Graph::AdjacentsTo and Graph::AdjacentsFrom functions](https://github.com/ignitionrobotics/ign-math/blob/5296217a8625994d07534f31551a899cf60a7a94/include/ignition/math/graph/Graph.hh#L291:383) return non-`const` references to the graph vertices, while the [Graph::IncidentsTo and Graph::IncidentsFrom functions](https://github.com/ignitionrobotics/ign-math/blob/5296217a8625994d07534f31551a899cf60a7a94/include/ignition/math/graph/Graph.hh#L425:486) only return `const` references to the graph edges.

# Steps to Reproduce

* Apply the following patch to [GraphUndirected_TEST.cc](https://github.com/ignitionrobotics/ign-math/blob/5296217a8625994d07534f31551a899cf60a7a94/src/graph/GraphUndirected_TEST.cc):

~~~
diff -r 5daad73583053b529a744f9939d5e5150f02c1f0 src/graph/GraphUndirected_TEST.cc
--- a/src/graph/GraphUndirected_TEST.cc	Tue Feb 12 20:49:41 2019 +0000
+++ b/src/graph/GraphUndirected_TEST.cc	Fri Feb 15 17:44:21 2019 -0800
@@ -365,9 +365,10 @@
   // Change some content and verity it.
   e2.Data() = 5.0;
   e2.SetWeight(6.0);
-  auto edge = graph.EdgeFromId(e2.Id());
+  auto &edge = graph.EdgeFromId(e2.Id());
   EXPECT_DOUBLE_EQ(5.0, edge.Data());
   EXPECT_DOUBLE_EQ(6.0, edge.Weight());
+  edge.Data() = 7.0;
 
   // Check that the edges point to the right vertices.
   EXPECT_EQ(0u, e0.Vertices().first);
~~~

* Try to build and it will fail to compile.

**Expected behavior:**  

Compile successfully.

**Actual behavior:**

Fails to compile.

**Reproduces how often:** 

always

# Versions

5.0.0 and 6.0.0, Ubuntu

# Additional Information

ms-jagadeeshan:
# 🦟 Bug fix

Fixes #99 

## Summary

Issue: The `Graph::VertexFromId` function has two versions that return a const and non-const reference to a Vertex. However there is only a const version of the `Graph::EdgeFromId` function, which makes it hard to modify Edge Data()

Solution : Have added non-const(mutable) reference function for `Graph::EdgeFromId` .

Issue: Likewise, the `Graph::AdjacentsTo` and `Graph::AdjacentsFrom` functions return non-const references to the graph vertices, while the `Graph::IncidentsTo` and `Graph::IncidentsFrom` functions only return const references to the graph edges.

Solution: Removed the `const` from return type of `Graph::IncidentsTo` and `Graph::IncidentsFrom` functions.


caguero (commented):
I think these changes break ABI (and API). We should retarget the patch to `main`.

Alternatively, we could add the functions that are missing without changing the existing ones.

ms-jagadeeshan:
@caguero Thank you for the guiding.
I have reverted back the breaking change, now the change only contain the new function.
Also the second point mentioned in the original issue is not clear, see this https://github.com/gazebosim/gz-math/issues/99#issuecomment-1972551077

caguero (approved):
The new function looks good to me.

PR #589, opened 2024-04-23
Add bzlmod support for gz-math8

mjcarroll:


iche033 (approved):
looks good to me, maybe @shameekganguly can take a look as well

shameekganguly (approved):
Thanks Michael!

include/gz/math/Matrix3.hh
```diff
@@ -21,9 +21,8 @@
 #include <cstring>
 #include <utility>
 #include <gz/math/Helpers.hh>
-#include <gz/math/Vector3.hh>
-#include <gz/math/Quaternion.hh>
```
azeey, line 25: Quaternion is used in this header, so I think we should keep this header. We can probably remove the forward declaration below actually.

PR #590, opened 2024-05-02
Enable 24.04 CI on harmonic

scpeters:
# 🎉 New feature

Attempting to enable CI on 24.04 for Harmonic (partial backport of #587)

## Summary

The python3-distutils package is not available on 24.04, but we only use it in a codepath for old versions of cmake. Since 24.04 has a new enough version of cmake, that package is moved from the generic `package.apt` file to distro-specific `packages-focal.apt` and `packages-jammy.apt` files.

* Don't install python3-distutils on 24.04
* Remove old ruby cmake code

## Test it

Observe CI results.

## Checklist
- [ ] Added tests

### DIFF
--- a/.github/ci/packages-jammy.apt
+++ b/.github/ci/packages-jammy.apt
@@ -1,1 +1,0 @@
-python3-distutils
--- a/.github/ci/packages-jammy.apt
+++ b/.github/ci/packages-jammy.apt
@@ -1,0 +1,1 @@
+python3-distutils
--- a/.github/ci/packages-focal.apt
+++ b/.github/ci/packages-focal.apt
@@ -1,0 +1,1 @@
+python3-distutils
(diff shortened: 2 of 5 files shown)


.github/ci/packages-jammy.apt
```diff
@@ -0,0 +1 @@
+python3-distutils
```
azeey, line 1: I don't think we need it in Jammy either (see https://github.com/gazebosim/gz-sim/pull/2374#discussion_r1572629902)

azeey (approved):

.github/ci/packages-jammy.apt
scpeters: you're right; I knew we need it for focal and wasn't thinking clearly about jammy. removed in https://github.com/gazebosim/gz-math/pull/590/commits/355780f8dd2fa40ec3d16c377b655406f84d1181

PR #596, opened 2024-05-26
Removed the workaround LOCAL2 coordinate system from SphericalCoordinates.

peci1:
# 🦟 Bug fix

Fixes https://github.com/gazebosim/gz-math/pull/235#discussion_r1442096703

## Summary

PR #235 introduced a workaround coordinate system `LOCAL2` that fixed a bug regarding the `LOCAL` system. This workaround was only intended to be kept for a tick-tock cycle, but it slipped from gz-math6 to gz-math7, too. This PR removes it from gz-math8.

All operations with the LOCAL coordinate system will now correctly use ENU as advertised.

Note that this breaks contract to code programmed for gz-math7 and earlier, where LOCAL system was actually WSU instead of ENU (rotated by 180 degrees heading).

There is one thing I'm not sure about. The `LOCAL2` system had index 5 in the enum. Now I removed it and we're again left with just `LOCAL` with value 4. Maybe it would be more practical to specify `LOCAL = 5` so that it is easier to spot cases in which users just supplied the value instead of the enum constant. But I'm not sure how probable that is.

This is a breaking change and will require downstream changes at least in https://github.com/gazebosim/gz-sim/blob/40aaddc7eb4ca12a03b64e18ffc101db9f9c24ec/src/Util.cc#L694 .

## Checklist
- [x] Added tests

### DIFF
--- a/Migration.md
+++ b/Migration.md
@@ -7,6 +7,19 @@ release will remove the deprecated code.
 
 ## Gazebo Math 8.X to 9.X
 
+### Additions
+
+1. **SphericalCoordinates.hh**
+    + `SphericalCoordinates::SphericalPositionFromLocal()`.
+    + `SphericalCoordinates::LocalPositionFromSpherical()`.
+    + `SphericalCoordinates::GlobalVelocityFromLocal()`.
+    + `SphericalCoordinates::LocalVelocityFromGlobal()`.
+    + Enum value `SphericalCoordinates::CoordinateType::LOCAL_TANGENT`
+      which substitutes the wrong `LOCAL` coordinate and improves the
+      name of `LOCAL2` (it is its functional equivalent).
+    + These functions use the newly introduced `LOCAL_TANGENT`
+      coordinate type instead of the wrong `LOCAL` frame.
+
 ### Deprecations
 
 1. **graph/Vertex.hh**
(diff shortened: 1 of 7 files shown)


scpeters (commented):
With this approach, downstream code using `LOCAL2` will fail to compile and code using `LOCAL` will have a behavior change that may not be discoverable unless they read the release notes. Breaking code is at least discoverable, but I'm concerned about silent behavior changes.

* If the goal is to get rid of the buggy behavior in a discoverable way, then we should remove the `LOCAL` enum.
* If we want a nicer name than `LOCAL2`, then we should make a new name like `LOCAL_Z_UP` or something and deprecate `LOCAL2` and/or print a warning when it is used.

What do you think?

peci1:
Thanks for the comments. This PR was meant as a kickoff for discussion.

I support the idea of removing LOCAL and renaming LOCAL2. LOCAL_Z_UP sounds doable, though not optimal. But I don't have anything better right now.

The question is what to do with SphericalFromLocalPosition. The best way I see would also be to rename it. Agree? E.g. SphericalFromLocalZUpPosition.

arjo129:
Hi @peci1, was wondering what the status on this was. Do we intend shipping this change with ionic (code freeze 26/8)? IF we can;t come to a consensus by then I will remove the beta label.

peci1:
I think it would be beneficial to have this in Ionic. I'll have a closer look at it next week.

peci1:
Okay, I'll try renaming LOCAL2 to LOCAL_TANGENT and deprecating LOCAL.

Regarding renaming of the functions, I don't think it's necessary as it will be obvious what they do.

peci1:
So, the PR is updated.

I've kept LOCAL and LOCAL2 and deprecated them. I've added LOCAL_TANGENT with the same numerical value as LOCAL.

I've also deprecated all `*From*(Position|Velocity)` and created `*(Position|Velocity)From*` instead. The new functions do the conversion using LOCAL_TANGENT, while the deprecated ones do them via LOCAL.

I've also added tests that verify the behavior of the deprecated functions so that some future updated doesn't break them before they're removed.

peci1:
Continued as a part of #616

PR #600, opened 2024-07-01
Deprecate math::clock alias

### LINKED ISSUE #436: Definition of math::clock
## Environment
* OS Version:  all
* Source or binary build? all
  
  

### Expected behavior

Maybe it's just me, but I was surprised when I noticed that there's an `ignition::math::clock` class, defined here:

https://github.com/gazebosim/gz-math/blob/49c3915175f371425702fe80e0240bc1b39eaaed/include/ignition/math/Stopwatch.hh#L31

and used for example here:

https://github.com/gazebosim/gz-math/blob/49c3915175f371425702fe80e0240bc1b39eaaed/src/Stopwatch_TEST.cc#L45-L46

I think it's dangerous to define a clock directly under the `math` namespace in a specific public header like this. It can collide with other clocks defined in other headers.
    
### Actual behavior: 

I'd expect that alias to be either:

* Internal to the `StopWatch` class and not visible through the public header
* Defined in a more global place like `Helpers.hh` if we want that to be the default clock type used throughout all `gz-math` classes.

## Steps to reproduce

Example on how to cause a compilation error:

```diff
diff --git a/include/ignition/math/Helpers.hh b/include/ignition/math/Helpers.hh
index 7152b522..35d83c1f 100644
--- a/include/ignition/math/Helpers.hh
+++ b/include/ignition/math/Helpers.hh
@@ -220,6 +220,7 @@ namespace ignition
   /// \brief Math classes and function useful in robot applications.
   namespace math
   {
+    using clock = std::chrono::system_clock;
     // Inline bracket to help doxygen filtering.
     inline namespace IGNITION_MATH_VERSION_NAMESPACE {
     //
diff --git a/src/Stopwatch_TEST.cc b/src/Stopwatch_TEST.cc
index 5df56321..2db84f37 100644
--- a/src/Stopwatch_TEST.cc
+++ b/src/Stopwatch_TEST.cc
@@ -18,6 +18,7 @@
 #include <gtest/gtest.h>
 #include <thread>
 
+#include "ignition/math/Helpers.hh"
 #include "ignition/math/Stopwatch.hh"
 
 using namespace ignition;
```

## Suggestion

At this point, removing that definition would break API. Considering that `steady_clock` is the only clock type used throughout the library (so far), we could consider moving that alias to `Helper.hh` so that it can be used by classes other than `StopWatch`, i.e. `SpeedLimiter`, `DiffDriveOdometry`, `GaussMarkovProcess`...

Alternatively, we could deprecate `ignition::math::clock` from version 7 and encourage downstream users to use `std::chrono` or define their own aliases as they see fit.

scpeters:
# 🎉 New feature

Closes #436.

## Summary

Recommend using `std::chrono::steady_clock` directly instead of this alias. This should already have been removed from downstream `gz-*` packages.

## Test it

Compile Ionic workspace with this branch and watch for deprecation warnings.

## Checklist
- [ ] Added tests
- [X] Updated migration guide (as needed)


include/gz/math/Stopwatch.hh
```diff
@@ -28,7 +28,9 @@ namespace gz
   namespace math
   {
     // Use a steady clock
-    using clock = std::chrono::steady_clock;
+    // This alias is now deprecated; please use std::chrono::steady_clock
+    // directly instead.
+    using clock GZ_DEPRECATED(8) = std::chrono::steady_clock;
```
scpeters, line 33: windows CI doesn't like this

~~~
Stopwatch.hh(33,17): error C2062: type 'int' unexpected
~~~

scpeters: trying `[[deprecated]]` based on https://stackoverflow.com/a/50882131

src/Stopwatch_TEST.cc
```diff
@@ -154,8 +155,8 @@ TEST(Stopwatch, StartStopReset)
   watch.Start(true);
   EXPECT_TRUE(watch.Running());
   EXPECT_LT(watch.StopTime(), watch.StartTime());
-  EXPECT_NE(math::clock::duration::zero(), watch.ElapsedRunTime());
-  EXPECT_EQ(math::clock::duration::zero(), watch.ElapsedStopTime());
+  EXPECT_NE(steady_clock::duration::zero(), watch.ElapsedRunTime());
```
scpeters, line 158: windows failed on this line; I think it's flaky

* [Build Status](https://build.osrfoundation.org/job/gz_math-pr-win/93/) https://build.osrfoundation.org/job/gz_math-pr-win/93/

scpeters: retrying:

* [Build Status](https://build.osrfoundation.org/job/gz_math-pr-win/95/) https://build.osrfoundation.org/job/gz_math-pr-win/95/

ahcorde (approved):

PR #603, opened 2024-07-01
Clean up nested namespaces

mjcarroll:
Massive diff is mostly whitespace as I'm condensing namespaces.

This is to be ported forward to main.


scpeters:
diff without whitespace: https://github.com/gazebosim/gz-math/pull/603/files?w=1

ahcorde (approved):

PR #605, opened 2024-07-02
Remove deprecated functions

mjcarroll:


scpeters:
alternative in https://github.com/gazebosim/gz-math/pull/601 with migration guide updates

PR #606, opened 2024-07-02
Fix potential unsafe initialization in the Graph class

### LINKED ISSUE #269: Unsafe Constant Initialization

caguero:
# 🦟 Bug fix

Partially fixes #269 (`Graph` class).

## Summary

I implemented the "Construct on first use" idiom (https://www.freecodecamp.org/news/cpp-static-initialization-order-fiasco/) to avoid the potential unsafe construction.

## Checklist
- [ ] Added tests


include/gz/math/graph/Graph.hh
```diff
@@ -150,7 +150,7 @@ namespace graph
         {
           std::cerr << "[Graph::AddVertex()] The limit of vertices has been "
                     << "reached. Ignoring vertex." << std::endl;
-          return Vertex<V>::NullVertex;
+          return NullVertex<V>();
```
scpeters, line 153: this seems like an API change; we may need to do it, but let me check if sdformat will be affected

scpeters: yes, this will require changes in sdformat:

https://github.com/gazebosim/sdformat/blob/main/src/FrameSemantics.cc#L89

include/gz/math/graph/Edge.hh
```diff
@@ -204,9 +205,6 @@ namespace graph
   template<typename E>
   class UndirectedEdge : public Edge<E>
   {
-    /// \brief An invalid undirected edge.
-    public: static UndirectedEdge<E> NullEdge;
```
scpeters, line 208: We may choose to make a hard API break here, but I'd prefer to not remove these static types in this pull request, but to deprecate them instead, so that we can coordinate fixes for sdformat and any other `gz-*` packages in separate pull requests.

scpeters (changes requested):
we should update the migration guide to document the API break / deprecation, depending on which route we take

caguero:
> we should update the migration guide to document the API break / deprecation, depending on which route we take

Migration guide updated. See [15b8975](https://github.com/gazebosim/gz-math/pull/606/commits/15b8975131bdef5f44c7932f4d120a15d21eca64) and [5f606f0](https://github.com/gazebosim/gz-math/pull/606/commits/5f606f0b6acfc11ec94ad077f7ea7229cd8ee44e).

include/gz/math/graph/Graph.hh
caguero: I restored the original static members and added deprecations. It should be possible to keep using them.
[15b8975](https://github.com/gazebosim/gz-math/pull/606/commits/15b8975131bdef5f44c7932f4d120a15d21eca64)

include/gz/math/graph/Edge.hh
caguero: Let me know if the changes in  [15b8975](https://github.com/gazebosim/gz-math/pull/606/commits/15b8975131bdef5f44c7932f4d120a15d21eca64) are sufficient to keep sdformat and friends compiling.

scpeters: it looks like sdformat15 will still compile with this change as of https://github.com/gazebosim/gz-math/pull/606/commits/5f606f0b6acfc11ec94ad077f7ea7229cd8ee44e (tested using the `ci_matching_branch/` trick partially documented in https://github.com/gazebosim/docs/pull/377)

[Build Status](https://build.osrfoundation.org/view/gz-ionic/job/sdformat15-install_bottle-homebrew-amd64/210/) https://build.osrfoundation.org/view/gz-ionic/job/sdformat15-install_bottle-homebrew-amd64/210/

Migration.md
```diff
@@ -5,6 +5,32 @@ Deprecated code produces compile-time warnings. These warning serve as
...
+
+1. **graph/Graph.hh**
+    + The functions `Graph::AddVertex()` and `Graph::VertexFromId()` might
+      return NullVertex() instead of NullVertex.
+      E.g.: https://github.com/gazebosim/gz-math/pull/606/files#diff-0c0220a7e72be70337975433eeddc3f5e072ade5cd80dfb1ac03da233c39c983L153-R153
+
+    + The functions `Graph::AddEdge()`, `Graph::LinkEdge()`,
+      `Graph::EdgeFromVertices()` and `Graph::EdgeFromId()` functions might
+      return NullEdge() instead of NullEdge.
+      E.g.: https://github.com/gazebosim/gz-math/pull/606/files#diff-0c0220a7e72be70337975433eeddc3f5e072ade5cd80dfb1ac03da233c39c983L222-R222
```
scpeters, line 20: I don't think these are breaking change in `Graph.hh`. I had originall commented on `Graph.hh` because that is where I noticed that our API was changing due to the previously removed type, but for this PR, I think the only needed change to the Migration guide is the deprecations below

Migration.md
```diff
@@ -5,6 +5,32 @@ Deprecated code produces compile-time warnings. These warning serve as
...
+
```
scpeters, line 21:
```suggestion
```

Migration.md
caguero: You're right, we're not changing the function signature so it should be fine.

include/gz/math/graph/Edge.hh
```diff
@@ -332,9 +336,19 @@ namespace graph
...
   DirectedEdge<E> DirectedEdge<E>::NullEdge(
     VertexId_P(kNullId, kNullId), E(), 1.0, kNullId);
+
+  /// \brief An invalid edge.
+  template<typename E, typename EdgeType>
+  EdgeType &NullEdge()
+  {
+    static auto e = std::make_unique<EdgeType>(
+      VertexId_P(kNullId, kNullId), E(), 1.0, kNullId);
+    return *e;
```
scpeters, line 350: looking at [Material::Predefined()](https://github.com/gazebosim/gz-math/blob/0ebc06e80adb5d3359957dd0d9b6b5e9d08013c2/src/Material.cc#L98-L119), I wonder if the unique pointer should be made and released from inside a static lambda function?

azeey: I think it would be better to use `gz::utils::NeverDestroyed` (e.g. https://github.com/gazebosim/sdformat/blob/f05f4e7ad1a6784f9ff1a6c1b362191677baa70d/src/Types.cc#L149). However, the fact that this function returns a non-const reference could be problematic since anyone can modify the value.

scpeters: regarding the non-const reference, I think the behavior in gz-math7 and earlier with static variables has a similar issue right? I tried changing the `NullEdge` and `NullVertex` functions to return a const reference, but lots of `Graph.hh` functions that currently return non-const references like to return references to `NullEdge` or `NullVertex` and fail to compile. This seems like a bigger issue with the API that we could open an issue about.

scpeters: trying again in https://github.com/gazebosim/gz-math/pull/612

PR #609, opened 2024-07-17
Revert "Fix potential unsafe initialization in the Graph class (#606)"

scpeters:
# 🦟 Bug fix

This reverts commit 17585a92918b16340d57b5746e0a50d8e114b2d4.

## Summary

It appears that #606 and https://github.com/gazebosim/sdformat/pull/1458 caused test failures in gz-sim (see https://github.com/osrf/buildfarm-tools/issues/67#issuecomment-2232120738). I opened a test PR in gz-sim (https://github.com/gazebosim/gz-sim/pull/2482) that builds against gz-math8 and sdformat15 with those changes reverted, and the tests were fixed:

[Build Status](https://build.osrfoundation.org/job/gz_sim-ci-pr_any-homebrew-amd64/637/) https://build.osrfoundation.org/job/gz_sim-ci-pr_any-homebrew-amd64/637/

I don't know why this broke tests, but I propose reverting these changes for now until we can figure it out.

## Checklist
- [X] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Rebase-and-Merge**.

PR #612, opened 2024-07-19
Retry fix for unsafe initialization in graph classes

### LINKED ISSUE #269: Unsafe Constant Initialization

scpeters:
# 🦟 Bug fix

Fixes #269.

## Summary

The is a second attempt at fixing the potentially unsafe initialization issues in the `graph::Edge` and `graph::Vertex` classes after the first attempt in #606 was reverted in #609 due to test failures that it caused in gz-sim (documented in https://github.com/osrf/buildfarm-tools/issues/67#issuecomment-2232120738). This pull request uses a static local variable instead of dynamic memory allocation and doesn't cause gz-sim test failures in my testing.

## Checklist
- [ ] Added tests
- [X] Updated migration guide (as needed)
- [X] Consider updating Python bindings (if the library has them)

### DIFF
--- a/include/gz/math/graph/Edge.hh
+++ b/include/gz/math/graph/Edge.hh
@@ -21,6 +21,7 @@
 #include <cstdint>
 #include <functional>
 #include <map>
+#include <memory>
 #include <ostream>
 #include <set>
 
@@ -203,7 +204,8 @@ namespace graph
   class UndirectedEdge : public Edge<E>
   {
     /// \brief An invalid undirected edge.
-    public: static UndirectedEdge<E> NullEdge;
+    // Deprecated in favor of NullEdge().
+    public: static UndirectedEdge<E> GZ_DEPRECATED(8) NullEdge;
 
     /// \brief Constructor.
     /// \param[in] _vertices The vertices of the edge.
@@ -255,6 +257,7 @@ namespace graph
   };
 
   /// \brief An invalid undirected edge.
+  // Deprecated in favor of NullEdge().
   template<typename E>
   UndirectedEdge<E> UndirectedEdge<E>::NullEdge(
     VertexId_P(kNullId, kNullId), E(), 1.0, kNullId);
@@ -266,7 +269,8 @@ namespace graph
   class DirectedEdge : public Edge<E>
   {
     /// \brief An invalid directed edge.
-    public: static DirectedEdge<E> NullEdge;
+    // Deprecated in favor of NullEdge().
+    public: static DirectedEdge<E> GZ_DEPRECATED(8) NullEdge;
 
     /// \brief Constructor.
     /// \param[in] _vertices The vertices of the edge.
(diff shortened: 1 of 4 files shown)


scpeters:
using the `ci_matching_branch/` technique with https://github.com/osrf/homebrew-simulation/commit/e682b2596e6cd40dc8be3c4a8414818a1c3f16ca and https://github.com/gazebosim/gz-sim/pull/2482, this branch and https://github.com/gazebosim/sdformat/pull/1460 are test in gz-sim CI:

* [Build Status](https://build.osrfoundation.org/job/gz_sim-ci-pr_any-homebrew-amd64/649/) https://build.osrfoundation.org/job/gz_sim-ci-pr_any-homebrew-amd64/649/

ahcorde (approved):

include/gz/math/graph/Edge.hh
```diff
@@ -330,9 +333,19 @@ namespace graph
...
+  // Deprecated in favor of NullEdge().
   template<typename E>
   DirectedEdge<E> DirectedEdge<E>::NullEdge(
     VertexId_P(kNullId, kNullId), E(), 1.0, kNullId);
+
+  /// \brief An invalid edge.
+  template<typename E, typename EdgeType>
+  EdgeType &NullEdge()
+  {
+    static EdgeType e(
```
azeey, line 345: This is still problematic depending the given types `E`, and `EdgeType` because the destructor eventually runs when the program exits. According to the [Google style guide](https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables), it's best to allocate memory from the heap and never delete it, which will avoid calling the destructor. `gz::utils::NeverDestroyed` does exactly that, so I recommend we switch to that.

azeey (changes requested):

include/gz/math/graph/Edge.hh
scpeters: trying `NeverDestroyed` in https://github.com/gazebosim/gz-math/pull/612/commits/3fa98549b1e781aca2f17909538c4891548963ad

scpeters: rerunning gz-sim CI [Build Status](https://build.osrfoundation.org/job/gz_sim-ci-pr_any-homebrew-amd64/693/) https://build.osrfoundation.org/job/gz_sim-ci-pr_any-homebrew-amd64/693/

azeey (approved):

Issue #200, opened 2021-03-25
Color: Invariant for channels within `[0, 1]` is not strictly enforced; implicity divison is weird


EricCousineau-TRI:
## Environment
* OS Version: N/A
* Source or binary build? Current master (see source permalink)

## Description
Came across this during @jennuine's resolution of https://github.com/osrf/sdformat/issues/193

* Expected behavior:
  * `Color` channel values are all strictly w/in `[0, 1]`
  * An error is thrown (rather than weird implicit conversion)

* Actual behavior: Here's what @seancurtis-tri found:
```
  // Case: Values out of range:
  //  A (alpha) simply gets clamped to the range [0, 1]
  //  For each individual element in R, G, B:
  //    Negative values are set to zero.
  //    Values > 1 are divided by 255
```
https://github.com/RobotLocomotion/drake/blob/v0.28.0/multibody/parsing/test/detail_scene_graph_test.cc#L861-L867

This is due to the following code:
https://github.com/ignitionrobotics/ign-math/blob/ad0118e689ba531c93df9c332e0c78a5dcd84e6e/src/Color.cc#L498-L511

## Steps to reproduce
See code

## Output
N/A

EricCousineau-TRI:
Per @jennuine's analysis, this invariant is now enforced in `libsdformat` XML parsing (https://github.com/osrf/sdformat/pull/519), but due to this issue, it can leak in via DOM mutators.

jwnimmer-tri:
What's even weirder is that the _copy constructor_ keeps dividing again on each copy, in case it was still over the limit:
https://github.com/ignitionrobotics/ign-math/blob/797f67f89c6863b5e02b6e7b24b6c587396cc614/src/Color.cc#L47-L51

Whichever of #283 or #286 merges will fix that part, though, by marked the copy constructor as defaulted.

PR #613, opened 2024-07-26
Remove clamping of color values

azeey:
# 🦟 Bug fix

Toward #200 

## Summary
This removes the clamping of color values, but it doesn't yet enforce the invariant for channels being within [0, 1]. The best way to do that would be to throw an exception, but we don't do that anywhere in the gz-math codebase.

## Checklist
- [ ] Added tests

### DIFF
--- a/include/gz/math/Color.hh
+++ b/include/gz/math/Color.hh
@@ -104,7 +104,6 @@ namespace gz::math
                             const float _a = 1.0)
     : r(_r), g(_g), b(_b), a(_a)
     {
-      this->Clamp();
     }
 
     /// \brief Copy Constructor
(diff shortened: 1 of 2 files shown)


ahcorde (approved):

azeey:
I've added enforcement of the color values range [0, 1] in https://github.com/gazebosim/gz-math/pull/613/commits/5f0ae54cdae221bcdf4f6640f4539cc252b4c4e3. Some of the test values needed to be updated because the clamping is different now: instead of dividing by 255 when a value is greater than 1, it's clamped to 1.

While I was trying to fix tests, I noticed that the functions that convert between RGB and YUV are not well defined. It's not clear what the input range is for YUV values. I also couldn't find a good reference for the equations used for the conversion. In fact, checking wikipedia, it looks like the math might be incorrect. As far as I can tell, these functions are not being used anywhere in Gazebo, so I propse we deprecate them.

iche033:
the YUV conversion code matches one of the formula in this post: https://www.vbforums.com/showthread.php?728455-What-is-the-CORRECT-conversion-for-YUV-to-RGB. 

the HSV conversion code seems to come from: https://stackoverflow.com/a/6930407

I tried a few online color converters using the values from our test and the results don't really match. I agree we should either fix the code or deprecate / remove it.

iche033 (approved):

azeey:
Yes, there seems to be many conventions around this and I don't think any of us are experts in this area. I'll deprecate it in a follow-up PR.

PR #615, opened 2024-08-06
Check for 0 stddev when generating a random number using normal distribution

iche033:
# 🦟 Bug fix

Fixes https://github.com/gazebosim/gz-sim/issues/2519

## Summary

We use [std::normal_distribution](https://en.cppreference.com/w/cpp/numeric/random/normal_distribution/normal_distribution) in our Rand class. From the doc:

```
The behavior is undefined if stddev is not greater than zero.
```

So added a check for 0 stddev (sigma) and return mean if its 0.

## Checklist
- [x] Added tests

### DIFF
--- a/src/Rand.cc
+++ b/src/Rand.cc
@@ -25,6 +25,8 @@
   #include <unistd.h>
 #endif
 
+#include <limits>
+
 #include "gz/math/Rand.hh"
 
 using namespace gz;
@@ -54,7 +56,8 @@ double Rand::DblUniform(double _min, double _max)
 //////////////////////////////////////////////////
 double Rand::DblNormal(double _mean, double _sigma)
 {
-  if (equal(_sigma, 0.0))
+  if (equal(_sigma, 0.0, std::numeric_limits<double>::epsilon()))
+
     return _mean;
 
   NormalRealDist d(_mean, _sigma);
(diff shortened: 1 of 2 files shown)


src/Rand.cc
```diff
@@ -54,6 +54,9 @@ double Rand::DblUniform(double _min, double _max)
 //////////////////////////////////////////////////
 double Rand::DblNormal(double _mean, double _sigma)
 {
+  if (equal(_sigma, 0.0))
```
azeey, line 57: The default epsilon for `equal` here is 1e-6, which means if the sigma is set to any smaller value, this branch would be taken. I suggest we either we do a direct comparison with `0.0` or set the epsilon value to [std::numeric_limits<T>::epsilon](https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon)

azeey (changes requested):

src/Rand.cc
iche033: updated to use `std::numeric_limits::epsilon` . [33b900f](https://github.com/gazebosim/gz-math/pull/615/commits/33b900fa9c923760bd274b89b17ed0a165492707)

azeey (approved):

PR #620, opened 2024-08-22
Fix tolerance for axis-angle conversion of quaternions

shameekganguly:
# 🦟 Bug fix

## Summary
Quaternions smaller than 1e-3 in length are currently treated as 0 when converting to axis angle. This is because the squared length of the quaternion is compared to 0 with an epsilon of 1e-6. The fix is to compare the length of the quaternion instead of the squared length so that only a quaternion of length smaller than 1e-6 (default epsilon in `math::equal(...)`) is treated as zero for this conversion.

## Checklist
- [x] Added tests

### DIFF
--- a/include/gz/math/Quaternion.hh
+++ b/include/gz/math/Quaternion.hh
@@ -543,8 +543,8 @@ namespace gz::math
     /// \param[out] _angle CCW angle in radians.
     public: void AxisAngle(Vector3<T> &_axis, T &_angle) const
     {
-      T len = sqrt(this->qx*this->qx + this->qy*this->qy + this->qz*this->qz);
-      if (equal<T>(len, static_cast<T>(0)))
+      T sqLen = this->qx*this->qx + this->qy*this->qy + this->qz*this->qz;
+      if (equal<T>(sqLen, static_cast<T>(0), 1e-12))
       {
         _angle = 0.0;
         _axis.Set(1, 0, 0);
@@ -552,7 +552,7 @@ namespace gz::math
       else
       {
         _angle = 2.0 * acos(this->qw);
-        T invLen =  1.0 / len;
+        T invLen =  1.0 / sqrt(sqLen);
         _axis.Set(this->qx*invLen, this->qy*invLen, this->qz*invLen);
       }
     }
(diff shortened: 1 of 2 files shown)


include/gz/math/Quaternion.hh
```diff
@@ -543,7 +543,7 @@ namespace gz::math
     /// \param[out] _angle CCW angle in radians.
     public: void AxisAngle(Vector3<T> &_axis, T &_angle) const
     {
-      T len = this->qx*this->qx + this->qy*this->qy + this->qz*this->qz;
+      T len = sqrt(this->qx*this->qx + this->qy*this->qy + this->qz*this->qz);
       if (equal<T>(len, static_cast<T>(0)))
```
scpeters, line 547:
```suggestion
      T len2 = this->qx*this->qx + this->qy*this->qy + this->qz*this->qz;
      if (equal<T>(len2, static_cast<T>(0), 1e-12))
```

Your proposed solution makes sense. An alternative implementation is to rename the `len` variable and pass `_epsilon=1e-12` to the [math::equal function in Helpers.hh](https://github.com/gazebosim/gz-math/blob/gz-math7_7.5.0/include/gz/math/Helpers.hh#L399-L407), which can avoids some `sqrt` calls. I'm fine either way, just wanted to mention this to see which you prefer.

scpeters (approved):

include/gz/math/Quaternion.hh
shameekganguly: Done: [a0dd4a8](https://github.com/gazebosim/gz-math/pull/620/commits/a0dd4a8a8c4cdc53a0b1ebe8d12558508481e842)

azeey:
There's a new compiler warning on macOS. The previous build (https://build.osrfoundation.org/job/gz_math-ci-pr_any-homebrew-amd64/170/) didn't have this. Not sure if that's because the OS versions are different or doing the `sqrt` as before avoids the warning.

shameekganguly:
> There's a new compiler warning on macOS. The previous build (https://build.osrfoundation.org/job/gz_math-ci-pr_any-homebrew-amd64/170/) didn't have this. Not sure if that's because the OS versions are different or doing the `sqrt` as before avoids the warning.

Looks like we have a template instantiation for Quaternion with integer data in https://github.com/gazebosim/gz-math/blob/gz-math7/src/python_pybind11/src/Quaternion.hh which caused this check to trip.
Added `static_cast` to T in [a406769](https://github.com/gazebosim/gz-math/pull/620/commits/a406769800f2f4664489394aaef7c97b4122b3dc)

PR #636, opened 2024-10-11
Permit building python bindings separately from gz-math library

scpeters:
# 🎉 New feature

Part of https://github.com/osrf/homebrew-simulation/issues/2834

## Summary

This allows the `src/python_pybind11/CMakeLists.txt` file to be built as a top-level cmake project against an external gz-math library. The first commit (https://github.com/gazebosim/gz-math/commit/646de60598cb728b1f8df1ab1ba9394bbd8b8879) is a patch from @traversaro [used in conda](https://github.com/conda-forge/gz-math-feedstock/blob/main/recipe/standalone_bindings.patch), and the second commit (https://github.com/gazebosim/gz-math/commit/d2bae803353fb929dbbc279fe003da77dfef000b) replaces the `GZ_LIB_INSTALL_DIR` cmake variable with `CMAKE_INSTALL_LIBDIR` (see [GzPackaging.cmake:111](https://github.com/gazebosim/gz-cmake/blob/gz-cmake4_4.0.0/cmake/GzPackaging.cmake#L111)) since `GZ_LIB_INSTALL_DIR` is not defined with the minimal cmake project added in https://github.com/gazebosim/gz-math/commit/646de60598cb728b1f8df1ab1ba9394bbd8b8879.

## Test it

1. Invoke cmake on the root `CMakeLists.txt` with using `-DSKIP_PYBIND11=ON` to build and install the gz-math library without python bindings.
2. Invoke cmake on `src/python_pybind11/CMakeLists.txt` with `-DPython_EXECUTABLE=/path/to/python` to build and install python bindings for a given python version
3. Repeat 2 with for each desired version of python

I have a draft of an updated homebrew formula using this branch in https://github.com/osrf/homebrew-simulation/pull/2836

## Checklist
- [X] Signed all commits for DCO
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

[...]

PR #639, opened 2024-10-20
Prototype: parse version number from package.xml

scpeters:
not for merging: part of #636

This uses a short python script to parse the version number from the package.xml file. It does minimal validation:

* ensuring that the root element is `<package>`
* takes the first child '<version>' element
* validate the text of the `<version>` element using the [regex from catkin_pkg](https://github.com/ros-infrastructure/catkin_pkg/blob/1.0.0/src/catkin_pkg/package_version.py#L55-L58)

In order to run this script from cmake, we need to find `Python3` before the `project()` statement. It also splits the version number and uses the major version in the project name.

If we like this approach, we can consider adding the script and a helper cmake function/macro to gz-cmake, so that the block of cmake code before the `project()` statement could be replaced by `find_package(gz-cmakeX)` and then the helper cmake call. Note that we can't wrap the `project()` call (from [cmake docs](https://cmake.org/cmake/help/latest/command/project.html#usage)):

> The top-level CMakeLists.txt file for a project must contain a literal, direct call to the project() command; loading one through the [include()](https://cmake.org/cmake/help/latest/command/include.html#command:include) command is not sufficient. If no such call exists, CMake will issue a warning and pretend there is a project(Project) at the top to enable the default languages (C and CXX).


scpeters:
superseded by https://github.com/gazebosim/gz-math/pull/642

PR #647, opened 2024-11-15
Simple test to do clang/asan builds on gz-math

mjcarroll:
Inspired by robotlocomotion/drake's approach

### DIFF
--- a/tools/BUILD.bazel
+++ b/tools/BUILD.bazel
@@ -1,0 +1,5 @@
+sh_binary(
+    name = "asan",
+    srcs = ["asan.sh"],
+    data = ["asan.supp"],
+)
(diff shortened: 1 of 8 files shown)


tools/bazel.rc
```diff
@@ -0,0 +1,36 @@
+common:clang --repo_env=CC=clang-15
+common:clang --repo_env=CXX=clang++-15
+build:clang --action_env=CC=clang-15
+build:clang --action_env=CXX=clang++-15
+build:clang --host_action_env=CC=clang-15
+build:clang --host_action_env=CXX=clang++-15
+
+### ASan build. Clang only. ###
```
shameekganguly, line 8: nit: consider adding that both `--config asan --config clang` need to be passed to bazel

tools/BUILD.bazel
```diff
@@ -0,0 +1,5 @@
+sh_binary(
```
shameekganguly, line 1: nit: consider adding a docstring on how this can be run locally
`bazel run --config asan --config clang tools:asan`?

shameekganguly (approved):

mjcarroll:
whoops.  @ahcorde I think I actually found a better way of doing this.  I'm going to revert.

PR #652, opened 2024-12-02
Change to globbed bazel targets

mjcarroll:
Draft until I can test downstream impacts


BUILD.bazel
```diff
@@ -36,1439 +36,66 @@ gz_configure_header(
...
-        ":Matrix4",
-        ":Plane",
-        ":Pose3",
-        "@gz-utils//:ImplPtr",
+    name = "math",
+    srcs = sources + private_headers,
+    hdrs = public_headers,
+    includes = [
+        "include",
+        "src",
```
shameekganguly, line 70: Do we need to add `src` to `includes`? I don't think any of the headers there should be publicly accessible?

shameekganguly (approved):
Thanks Michael!

eigen3/BUILD.bazel
```diff
@@ -13,53 +13,52 @@ license(
...
 )
 
-cc_test(
-    name = "Conversions_TEST",
-    srcs = ["src/Conversions_TEST.cc"],
-    deps = [
-        ":Conversions",
-        "@googletest//:gtest",
-        "@googletest//:gtest_main",
+private_headers = glob(
```
shameekganguly, line 22: Bazel 8 doesn't seem to like the empty glob:

ERROR: Traceback (most recent call last):
	File "/home/runner/work/gz-math/gz-math/eigen3/BUILD.bazel", line 22, column 23, in <toplevel>
		private_headers = glob(
Error in glob: glob pattern 'src/*.hh' didn't match anything, but allow_empty is set to False (the default value of allow_empty can be set with --incompatible_disallow_empty_glob).
ERROR: package contains errors: eigen3

Should we just remove the `private_headers` list from this file?

BUILD.bazel
```diff
@@ -36,1458 +36,67 @@ gz_configure_header(
...
-cc_library(
-    name = "Error",
-    srcs = ["src/Error.cc"],
-    hdrs = ["include/gz/math/detail/Error.hh"],
-    includes = ["include"],
-    deps = [
-        ":Config",
-        ":Export",
+    exclude = [
+        "include/gz/math/sdf.hh",
```
shameekganguly, line 46: Maybe I'm not looking in the right place, but I don't see this file anywhere?

mjcarroll: Ah, copy pasted from sdformat bazel rules, will remove.

examples/BUILD.bazel
```diff
@@ -9,154 +9,16 @@ license(
...
-
-cc_binary(
-    name = "helpers",
-    srcs = ["helpers_example.cc"],
-    deps = [
-        "//:Helpers",
-        "//:Quaternion",
+example_sources = glob(
+    include = [
+        "*_example.cc",
```
shameekganguly, line 14: Not all of the examples seem to follow the "*_example.cc" naming convention. Should this glob be just "*.cc" instead?

shameekganguly: Actually, I think we can maintain parity with CMake here and keep the explicit build targets: [examples/CMakeLists.txt](https://github.com/gazebosim/gz-math/blob/gz-math8/examples/CMakeLists.txt)

shameekganguly (approved):

examples/BUILD.bazel
mjcarroll: Done.

scpeters:
it looks like sdformat bazel CI has been failing since this was merged (see https://github.com/gazebosim/sdformat/pull/1544#issuecomment-2655476811)

https://github.com/gazebosim/sdformat/pull/1540 sounds similar, but it fails too

PR #656, opened 2024-12-30
Encapsulate std::cerr under a single point of control

jwnimmer-tri:
(This isn't necessarily a new feature, but there doesn't seem to be a "cleanup" or "refactor" PR template.)

# 🎉 New feature

Closes (n/a).

## Summary

When integrating this library into a larger application, downstream code doesn't always want to spew onto cerr. To allow customization, we'll refactor to encapsulate the destination inside of a function. Future work can extend this with a mechanism to redirect it (e.g., to send it to spdlog instead of cerr, or to throw an exception instead of returning wrong answers).

## Test it

The existing unit tests provide the same level of code-coverage for error printing as previously (e.g., demonstrating that the error printing does not crash).

## Checklist
- [x] (N/A) Added tests
- [x] (N/A) Added example and/or tutorial
- [x] (N/A) Updated documentation (as needed)
- [x] (N/A) Updated migration guide (as needed)
- [x] (N/A) Consider updating Python bindings (if the library has them)

### DIFF
--- a/include/gz/math/PiecewiseScalarField3.hh
+++ b/include/gz/math/PiecewiseScalarField3.hh
@@ -18,14 +18,15 @@
 #define GZ_MATH_PIECEWISE_SCALAR_FIELD3_HH_
 
 #include <algorithm>
-#include <iostream>
 #include <limits>
+#include <sstream>
 #include <utility>
 #include <vector>
 
 #include <gz/math/Region3.hh>
 #include <gz/math/Vector3.hh>
 #include <gz/math/config.hh>
+#include <gz/math/detail/Error.hh>
 
 namespace gz::math
 {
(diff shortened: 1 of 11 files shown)


src/Error.cc
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+#include <iostream>
+
+#include "gz/math/detail/Error.hh"
+
+void gz::math::detail::LogErrorMessage(const std::string& message) {
```
ahcorde, line 21:
```suggestion
void gz::math::detail::LogErrorMessage(const std::string& message)
{
```

ahcorde (changes requested):
just a minor change, otherwise look good to me

ahcorde (approved):

include/gz/math/PiecewiseScalarField3.hh
```diff
@@ -77,21 +78,23 @@ namespace gz::math
       {
         if (pieces[i].region.Empty())
         {
-          std::cerr << "Region #" << i << " (" << pieces[i].region
-                    << ") in piecewise scalar field definition is empty."
-                    << std::endl;
+          std::ostringstream error_str;
```
azeey, line 81: nit: variables should be camelCase.

azeey (approved):
Thanks for the PR. FYI, we have recently added spdlog based logging in gz-utils (https://github.com/gazebosim/gz-utils/pull/134). The long term plan is to replace all console logging here in gz-math with the that.

jwnimmer-tri:
I pushed a rename from `error_str` to `errStream`.

---

FYI on spdlog --

I'm less enamored with spdlog than I was years ago.  I don't really think it is API and ABI-stable enough to be used a public (header file) dependency in low-level libraries like this one, especially those that try themselves to be ABI-stable.  I especially recommend against using spdlog in header files -- I'm going to be working on removing it from Drake header files this year.  I realize you can pin it to the distro or something to improve stability, but to me it seems like a bridge too far.  My recommendation is to limit the use of spdlog to just the one new `Error.cc` implementation file which can trampoline the message text into a spdlog logger.  (Then, we will still have a single point of control to replace error printing with exception raising.)  That's the design I'll be using in Drake.

If you want to work improving the error architecture in gz-math, my advice would actually be to stop printing messages entirely (in a future major release that has API changes).  Functions that receive bad input shouldn't log and proceed anyway, they should either throw an exception with the message, or return a `std::expected<T, E>` wrapper which can denote the return type xor error.  Low level libraries like gz-math that hide errors from their callers (by only ever logging them, instead of returning them) are extremely frustrating when we try to use them as building blocks for robust, higher-layer software.  In that light, I don't think gz-math really needs _any_ kind of logging whatsoever, so introducing spdlog down here is moving in the wrong direction.

PR #661, opened 2025-02-10
bazel: Build pybind module and test

mjcarroll:


MODULE.bazel
```diff
@@ -17,3 +20,34 @@ archive_override(
     strip_prefix = "gz-utils-gz-utils3",
     urls = ["https://github.com/gazebosim/gz-utils/archive/refs/heads/gz-utils3.tar.gz"],
 )
+
+PYTHON_VERSIONS = [
+    "3.8",
```
sloretz, line 25: FYI Python 3.8 is EOL https://devguide.python.org/versions/

BUILD.bazel
```diff
@@ -36,1458 +38,102 @@ gz_configure_header(
...
-    name = "Frustum_TEST",
-    srcs = ["src/Frustum_TEST.cc"],
-    deps = [
-        ":Frustum",
-        ":Helpers",
-        "@googletest//:gtest",
-        "@googletest//:gtest_main",
+    name = "gz-math",
+    srcs = sources,
+    hdrs = public_headers + private_headers,
```
sloretz, line 70: Private headers usually go in `srcs` instead.

> For cc_library rules, headers in hdrs comprise the public interface of the library and can be directly included both from the files in hdrs and srcs of the library itself as well as from files in hdrs and srcs of cc_* rules that list the library in their deps. Headers in srcs must only be directly included from the files in hdrs and srcs of the library itself. When deciding whether to put a header into hdrs or srcs, you should ask whether you want consumers of this library to be able to directly include it. This is roughly the same decision as between public and private visibility in programming languages.

https://bazel.build/reference/be/c-cpp#hdrs

MODULE.bazel
mjcarroll: removed

scpeters:
conflicts

mjcarroll:
> conflicts

Resolved.

scpeters (approved):
looks ok to my untrained eye

@sloretz any additional thoughts?

MODULE.bazel
```diff
@@ -17,3 +20,33 @@ archive_override(
...
+    for python_version in PYTHON_VERSIONS
+]
+
+pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
+
+[
+    pip.parse(
+        hub_name = "xacro_python_dependencies",
+        python_version = python_version,
+        requirements_lock = "//bazel:requirements_lock.txt",
```
sloretz, line 47: It looks like `//bazel:requirements_lock.txt` is missing from this PR

mjcarroll: Removed.

PR #784, opened 2026-07-06
Enable caching in bazel workflow

scpeters:
# 🦟 Bug fix

Part of https://github.com/gazebo-tooling/release-tools/issues/1518.

## Summary

This upgrades to the latest version of the bazel workflow, which has some fixes for caching which should speed up CI:

* https://github.com/bazel-contrib/.github/pull/59
* https://github.com/bazel-contrib/.github/pull/60

The bazel workflow ran once when the pull request was opened, and then I reran it to check for the effect of caching:

* [first invocation](https://github.com/gazebosim/gz-math/actions/runs/28828440158/attempts/1) completed in 3m51s
* [second invocation](https://github.com/gazebosim/gz-math/actions/runs/28828440158) completed in 1m19s

## Backport Policy

- [x] This is safe to backport to the following versions:
    - [x] Jetty
    - [x] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] This **should not** be backported
- [ ] I am not sure
- [ ] Other (fill in yourself)

## Checklist
- [x] Signed all commits for DCO
- [ ] Added a screen capture or video to the PR description that demonstrates the fix (as needed)
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] Updated Bazel files (if adding new files). Created an issue otherwise.
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
- [ ] Was GenAI used to generate this PR? If so, make sure to add "Generated-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` and `Generated-by` messages.

**Backports:** If this is a backport, please use **Rebase and Merge** instead.

PR #796, opened 2026-08-18
bazel.yml: experiment with cache_version

scpeters:
# 🎉 New feature

Part of https://github.com/gazebo-tooling/release-tools/issues/1535.

## Summary

Try setting `cache_version` to see if this improves cache hit rate. I believe it needs to be merged before it takes effect.

## Backport Policy

- [ ] This is safe to backport to the following versions:
    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [x] This **should not** be backported
- [ ] I am not sure
- [ ] Other (fill in yourself)

## Test it

Confirm that Bazel CI still passes. Won't be able to fully evaluate until it has landed on `main` and other CI jobs have had a chance to run.

## Checklist
- [ ] Added a screen capture or video to the PR description that demonstrates the feature
- [ ] Added tests
- [ ] Updated Bazel files (if adding new files). Created an issue otherwise.
- [ ] Was GenAI used to generate this PR? If so, make sure to add "Assisted-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` and `Generated-by` messages.

**Backports:** If this is a backport, please use **Rebase and Merge** instead.

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -14,6 +14,7 @@ jobs:
   test:
     uses: bazel-contrib/.github/.github/workflows/bazel.yaml@v7.7.0
     with:
+      cache_version: rotary
       folders: |
         [
           "."


shameekganguly:
Gemini suggests that the cache_version parameter should only be updated for invalidating caches rather than to improve cache hit rate: https://paste.googleplex.com/6563674448855040 (sorry internal access only, but I've spelled out the summary)

>After looking into how the `cache_version` parameter interacts with GitHub Actions and Bazel, and I don't think we should make this change.
In a nominal "happy path," GitHub Actions already scopes caches to the specific branch, so distinct cache versions don't actually change anything. However, if a cache gets evicted on a release branch, GitHub Actions falls back to downloading the cache from main.
If we merge this and change the `cache_version` strings per branch, we intentionally sever that fallback link. While our first-party code might have diverged between branches, Bazel caches at the action level. This means falling back to main's cache still gives us massive cache hits for heavy external dependencies (like protobuf, gtest, etc.) that haven't changed.
If we isolate the caches with unique version strings, an evicted cache on a release branch would force a 100% clean build of the entire third-party dependency tree from scratch.
Because of the huge performance benefit of sharing dependency caches across branches, we should keep the cache versions unified (or blank). We really only need to touch `cache_version` if we have a corrupted cache that we need to manually bust.

scpeters:
ok, thanks @shameekganguly 

I see the bazel actions take around 3-4 minutes pretty consistently for this repo, and I was guessing that there was room for improvement in the caching, but it doesn't sound like this will provide it

shameekganguly:
@scpeters looks like we haven't backported https://github.com/gazebosim/gz-math/pull/784 to the Jetty and Ionic branches, which would enable caching on those branches

shameekganguly:
Also, looking at the bazel CI workflow runs on `main` since bumping to version 7.7.0 of the workflow (which enabled caching), the workflow runs are further apart than Github's [7-day cache retention policy](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#default-limits). As a result, we haven't been able to take advantage of caching yet: https://github.com/gazebosim/gz-math/actions/workflows/bazel.yml?query=branch%3Amain

shameekganguly:
On the other hand, for `gz-sim`, it seems we are constantly at the cache limit, https://github.com/gazebosim/gz-sim/actions/caches, occasionally causing cache misses on `main`.

scpeters:
> Also, looking at the bazel CI workflow runs on `main` since bumping to version 7.7.0 of the workflow (which enabled caching), the workflow runs are further apart than Github's [7-day cache retention policy](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#default-limits). As a result, we haven't been able to take advantage of caching yet: https://github.com/gazebosim/gz-math/actions/workflows/bazel.yml?query=branch%3Amain

ok, that makes a lot of sense. My goal is to create a scheduled CI job that runs tests repeatedly looking for flakes, so if we schedule it to run more frequently than every 7 days, I would expect it to take advantage of the caching.

thanks @shameekganguly!

PR #814, opened 2026-08-21
Line3: use T instead of double in APIs

scpeters:
# 🎉 New feature

Similar to #809 

## Summary

I noticed that `Line3` is a templated class but still uses `double` in many API arguments. This updates the Constructor and Set APIs to use `T` more consistently instead of double.

Since these are template APIs, it might be safe to backport, but I'm not sure.

## Backport Policy

- [ ] This is safe to backport to the following versions:
    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] This **should not** be backported
- [x] I am not sure
- [ ] Other (fill in yourself)

## Test it

I don't see any uses of this class outside of gz-math, so verifying that CI passes should be sufficient

## Checklist
- [x] Signed all commits for DCO
- [ ] Added a screen capture or video to the PR description that demonstrates the feature
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [x] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] Updated Bazel files (if adding new files). Created an issue otherwise.
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
- [x] Was GenAI used to generate this PR? If so, make sure to add "Assisted-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

Assisted-by: Gemini 3.6 Flash

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` and `Generated-by` messages.

**Backports:** If this is a backport, please use **Rebase and Merge** instead.

PR #813, opened 2026-08-21
Line3: increase test coverage

scpeters:
# 🦟 Bug fix

Improve test coverage of the `Line3` class, similar to #807 

## Summary

This increases test coverage based on suggestions from Gemini and uses `static_cast<T>` where needed to fix windows C4244 compiler warnings.

### Summary of test coverage improvements

1. Default Constructor: Tested `Line3()` initialization.                                                                                           
2. Set Overload: Added tests for `Set(const Vector3<T>&, const Vector3<T>&)`.                                                                      
3. Intersect Edge Cases: Added tests for parallel non-overlapping line segments, parallel overlapping segments, and non-intersecting skew lines. 
4. Template Types: Added test cases for `Line3i` (int) and `Line3f` (float).

## Backport Policy

- [x] This is safe to backport to the following versions:
    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [x] Fortress
- [ ] This **should not** be backported
- [ ] I am not sure
- [ ] Other (fill in yourself)

## Checklist
- [x] Signed all commits for DCO
- [ ] Added a screen capture or video to the PR description that demonstrates the fix (as needed)
- [x] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] Updated Bazel files (if adding new files). Created an issue otherwise.
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
- [x] Was GenAI used to generate this PR? If so, make sure to add "Assisted-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

Assisted-by: Gemini 3.6 Flash

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` and `Generated-by` messages.

**Backports:** If this is a backport, please use **Rebase and Merge** instead.

PR #818, opened 2026-08-21
Line3: increase test coverage and use T instead of double in APIs (#814) (#813)

ahcorde:
backport https://github.com/gazebosim/gz-math/pull/813 and https://github.com/gazebosim/gz-math/pull/814


scpeters:
I will backport these individually

PR #819, opened 2026-08-21
Line3: increase test coverage and use T instead of double in APIs (#814) (#813)- #818

ahcorde:
backport https://github.com/gazebosim/gz-math/pull/813 and https://github.com/gazebosim/gz-math/pull/814


scpeters:
I would like to get a second review on whether it's safe to API to backport https://github.com/gazebosim/gz-math/pull/814, given that it is a templated class

scpeters:
thanks for assisting with the backports, but I'd prefer to backport them individually. I will close these and use mergify

PR #809, opened 2026-08-21
Line2: use T instead of double in APIs

scpeters:
# 🎉 New feature

Fixes some windows compiler warnings noticed in #804 and https://github.com/gazebosim/gz-math/pull/807

## Summary

While testing support for python bindings on Windows in #804, I noticed a bunch of C4244 compiler warnings from `Line2.hh`:

> conversion from 'double' to 'T', possible loss of data

Then I noticed that `Line2` is a templated class but still uses `double` in many API arguments. This updates the API to use `T` more consistently. There are still some instances of `double` arguments in the `Slope` function and for tolerance arguments, so `static_cast`s are added to handle these type conversions.

Summary from Gemini edited by me:

1. Constructor & Set arguments changed from double to T.
2. CrossProduct return type changed from double to T.
3. Updated Collinear, Parallel, Within, and Intersect to cast zero constants and _epsilon to T, avoiding mixed double/T arithmetic.

Since these are template APIs, it might be safe to backport, but I'm not sure.

## Backport Policy

- [ ] This is safe to backport to the following versions:
    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] This **should not** be backported
- [x] I am not sure
- [ ] Other (fill in yourself)

## Test it

I don't see any uses of this class outside of gz-math, so verifying that CI passes should be sufficient

## Checklist
- [x] Signed all commits for DCO
- [ ] Added a screen capture or video to the PR description that demonstrates the feature
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [x] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] Updated Bazel files (if adding new files). Created an issue otherwise.
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
[...]

PR #833, opened 2026-08-25
Revert "Line2: use T instead of double in APIs (backport #809)"

ahcorde:
Reverts gazebosim/gz-math#830


ahcorde:
@scpeters I think you are right, we can't backport this one

scpeters:
I was just chatting with @caguero and he thinks it's ok because it's a templated class

caguero:
I think the original PR was fine to backport.

PR #847, opened 2026-09-11
Fix volumeBelow and CenterOfVolumeBelow instability

ahcorde:
# 🦟 Bug fix

## Summary

  - What was wrong: in include/gz/math/detail/Box.hh, VolumeBelow and CenterOfVolumeBelow use exact formulas that subtract nearly equal numbers. When the plane is almost parallel to a face,
    those subtractions lose all precision. The formulas came in with gazebosim/gz-math#724.
  - What I changed:
    - An axis the plane is parallel to within rounding is now treated like an exactly parallel one, with the plane shifted by that axis's average contribution.
    - The centre along that axis gets an exact first-order correction, which reproduces the physical offset −θL²/(12D), so a tilted hull keeps its small restoring moment.
    - The cutoff scales with the float type's epsilon (about 1e-4 for double, 2e-2 for float). Integer boxes behave exactly as before.
  - Accuracy, measured against exact prism formulas for tilts from 0 to 1e-3, with one or two tiny slopes:

  |               | Volume error (relative) | Centre error                         |
  |---------------|-------------------------|--------------------------------------|
  | Before        | up to 2×10⁹             | up to 6×10¹³ m, some centres missing |
  | After, double | 7×10⁻¹⁴                 | under 6×10⁻¹⁰ m                      |
  | After, float  | 1.5×10⁻⁷                | 1.7×10⁻⁶ m                           |

## Backport Policy

- [x] This is safe to backport to the following versions:
    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [x] Fortress
- [ ] This **should not** be backported
- [ ] I am not sure
- [ ] Other (fill in yourself)

## Checklist
- [ ] Added a screen capture or video to the PR description that demonstrates the fix (as needed)
- [x] Added tests
- [ ] Updated Bazel files (if adding new files). Created an issue otherwise.
- [x] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
- [x] Was GenAI used to generate this PR? If so, make sure to add "Assisted-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

Assisted-by: Claude Opues 4.8

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` and `Generated-by` messages.

**Backports:** If this is a backport, please use **Rebase and Merge** instead.

### DIFF
--- a/include/gz/math/detail/Box.hh
+++ b/include/gz/math/detail/Box.hh
@@ -22,11 +22,31 @@
 #include <algorithm>
 #include <array>
 #include <cmath>
+#include <limits>
 #include <optional>
 #include <vector>
 
 namespace gz::math
 {
+namespace detail
+{
+/// \brief Smallest M_i = |n_i| * size_i that the Box inclusion-exclusion
+/// formulas treat as an axis the plane crosses, rather than one it is
+/// parallel to. Differencing F(x) - F(x - M_i) cancels catastrophically as
+/// M_i shrinks, leaving the centre of volume with a relative error near
+/// eps * (Msum / M_i)^2. Dropping the axis instead errs by about
+/// (M_i / Msum)^2, so the cut sits where the two meet, eps^(1/4) * Msum, and
+/// both stay near sqrt(eps).
+/// \param[in] _msum Sum of M_i over the three axes.
+/// \return The threshold. Zero for integer types, which do not round.
+template<typename T>
+T BoxNegligibleSpan(T _msum)
+{
+  return static_cast<T>(_msum *
+      std::sqrt(std::sqrt(std::numeric_limits<T>::epsilon())));
+}
+}  // namespace detail
+
 //////////////////////////////////////////////////
 template<typename T>
 Box<T>::Box(T _length, T _width, T _height)
@@ -159,8 +179,32 @@ T Box<T>::VolumeBelow(const Plane<T> &_plane) const
   if (alpha >= Msum)
     return totalVol;
 
-  // Count how many M_i are > 0
-  int nonzero = (M1 > 0 ? 1 : 0) + (M2 > 0 ? 1 : 0) + (M3 > 0 ? 1 : 0);
+  // Keep the axes whose M_i is not negligible. A negligible one is an axis
+  // the plane is parallel to within rounding: drop it as if M_i were zero,
+  // moving the plane by the mean of its contribution, M_i / 2, which keeps
+  // the volume accurate to second order in M_i. See BoxNegligibleSpan.
+  const T negligible = detail::BoxNegligibleSpan(Msum);
+  std::array<T, 3> Mv{};
+  int nonzero = 0;
+  T alphaKept = alpha;
+  T keptSum = 0;
+  for (const T Mi : {M1, M2, M3})
+  {
+    if (Mi > negligible)
+    {
+      Mv[nonzero++] = Mi;
+      keptSum += Mi;
+    }
+    else
+    {
+      alphaKept -= Mi / 2;
+    }
+  }
+
+  if (alphaKept <= 0)
+    return 0;
+  if (alphaKept >= keptSum)
+    return totalVol;
 
   auto cube = [](T x) -> T { return x * x * x; };
   auto clampPos = [](T x) -> T { return x > 0 ? x : 0; };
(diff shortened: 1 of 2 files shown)


caguero (commented):
Thanks for catching this! I just left a few small comments.

include/gz/math/detail/Box.hh
```diff
@@ -22,11 +22,39 @@
...
+inline namespace GZ_MATH_VERSION_NAMESPACE {
+namespace detail
+{
+/// \brief Smallest M_i = |n_i| * size_i that the Box inclusion-exclusion
+/// formulas treat as an axis the plane crosses, rather than one it is
+/// parallel to. Differencing F(x) - F(x - M_i) cancels catastrophically as
+/// M_i shrinks, leaving the centre of volume with a relative error near
+/// eps * (Msum / M_i)^2. Dropping the axis instead errs by about
+/// (M_i / Msum)^2, so the cut sits where the two meet, eps^(1/4) * Msum, and
+/// both stay near sqrt(eps).
```
caguero: This estimate holds when only one axis is small. With two axes both just above the threshold, the rounding error of the exact formula grows like eps * Msum^4 / (M_i^2 * M_j * M_k). For example normal (2e-5, 1.4e-4, 1) with offset 0.07 on the test box gives a centre error of about 3.7e-6 m in double. Still tiny, but it might be worth tweaking the claim here so nobody tunes the threshold from it later.

include/gz/math/detail/Box.hh
```diff
@@ -22,11 +22,39 @@
 #include <algorithm>
 #include <array>
 #include <cmath>
+#include <limits>
```
caguero, line 25: Nit:  Add `<initializer_list>`?

include/gz/math/detail/Box.hh
```diff
@@ -159,52 +187,70 @@ T Box<T>::VolumeBelow(const Plane<T> &_plane) const
...
   {
     // 1D case
-    T Mk = M1 > 0 ? M1 : (M2 > 0 ? M2 : M3);
-    T frac = alpha / Mk;
+    T frac = alphaKept / Mv[0];
     return totalVol * std::clamp(frac, T(0), T(1));
   }
   else
   {
     // 0D case: degenerate box
```
caguero: I think this branch is unreachable now. If nothing is kept then `keptSum` is 0, and the two early returns above already cover `alphaKept <= 0` and `alphaKept >= 0`. Same for the `k == 0` check in `CenterOfVolumeBelow`. I think we can drop both.

include/gz/math/detail/Box.hh
```diff
@@ -269,22 +317,43 @@ std::optional<Vector3<T>>
   const std::array<T, 3> half = {h1, h2, h3};
   const std::array<T, 3> nComp = {n.X(), n.Y(), n.Z()};
 
-  // Identify non-trivial axes (where M_i > 0)
+  // Identify non-trivial axes: those whose M_i is not negligible. A
```
caguero: Optional: this loop, the two early returns and the alpha / M / Msum setup are now duplicated between the two methods. A small helper in `detail` that returns the kept axes, `alphaKept` and `keptSum` would keep them from drifting apart.

src/Box_TEST.cc
```diff
@@ -17,10 +17,68 @@
...
+      for (const T d : {T(-0.05), T(0), T(0.035)})
+      {
+        for (const T sign : {T(1), T(-1)})
+        {
+          const double px = sign * tx;
+          const double py = ty;
+          const double depth = d + height / 2.0;
+          const double lengthSq = static_cast<double>(length) * length;
+          const double widthSq = static_cast<double>(width) * width;
+          SCOPED_TRACE("tx " + std::to_string(px) + " ty " +
```
caguero: Nit: `std::to_string` prints 1e-12 as `0.000000`, so the trace will not tell the slopes apart on a failure. A `std::ostringstream` or `testing::Message()` would keep the exponent.

caguero (approved):

ahcorde:
https://github.com/Mergifyio backport gz-math9 gz-math8 gz-math7 ign-math6

